PR #66, opened 2016-03-29
Feedstocks: Warn user about empty repos

jakirkham:
Fixes https://github.com/conda-forge/conda-forge.github.io/issues/54

In some cases, feedstock generation fails. When this happens, repos could be created, but have not commits in them. As a result, we can't add them as a submodule because we don't have a commit to point to. This resulted in a failure. However, the error message provided was a bit uninformative for those unfamiliar with the code. Also, it stops adding other working repos. As there may be cases where we have a broken repo and we don't want this grinding to a halt, here we raise a warning that both explains what is going on and let's us proceed through the working repos.

### DIFF
--- a/scripts/update_feedstocks_submodules.py
+++ b/scripts/update_feedstocks_submodules.py
@@ -16,6 +16,8 @@ from git import Repo
 
 import argparse
 
+import warning
+
 parser = argparse.ArgumentParser(description='Update the conda-forge/feedstocks repo.')
 parser.add_argument('feedstocks_repo', help="The location of the checked out conda-forge/feedstocks repo.")
 
@@ -40,8 +42,16 @@ for feedstock in forge_feedstocks:
             feedstocks_repo.submodules[feedstock.package_name].remove()
 
         # Add the new submodule.
-        feedstocks_repo.create_submodule(feedstock.package_name, repo_subdir,
-                                         url=feedstock.clone_url)
+        try:
+            feedstocks_repo.create_submodule(feedstock.package_name, repo_subdir,
+                                             url=feedstock.clone_url)
+        except ValueError:
+            warning.warn(
+                    "Unable to add the submodule {}. "
+                    "This is likely because the repo has no commits, "
+                    "which likely means something went wrong with feedstock generation. "
+                    "Will skip adding this submodule and continue.".format(feedstock.package_name)
+            )
 
 
 # Pick out the feedstocks which exist on the repo, but are no longer on conda-forge.


jakirkham:
cc @ocefpaf

scripts/update_feedstocks_submodules.py
```diff
@@ -40,8 +42,16 @@
...
+        try:
+            feedstocks_repo.create_submodule(feedstock.package_name, repo_subdir,
+                                             url=feedstock.clone_url)
+        except ValueError:
+            warning.warn(
+                    "Unable to add the submodule {}. "
+                    "This is likely because the repo has no commits, "
+                    "which likely means something went wrong with feedstock generation. "
+                    "Will skip adding this submodule and continue.".format(feedstock.package_name)
+            )
```
ocefpaf: :+1: for the message, but should that still raise an error instead of just warning?

jakirkham: I don't thinks so because if it raises an error we can't process the rest of the working feedstocks.

In this case, I am looking at yt, but there will be more until we can get the issue with `libmo_unpack` fixed. Also, I doubt this is the last time that we are going to want to bypass broken feedstocks.

The main problem is less that they get added to the feedstocks repo and more that permissions to feedstocks aren't being handed out to people in a timely manner when these issues occur.

ocefpaf: I am not convinced that passing this with just a warning is the way to go but I understand your point. I guess that @pelson is out and he cannot review https://github.com/conda-forge/conda-smithy/pull/106 right today. So go ahead and merge this change. If it turns out to be a bad idea we can just reverse it later.

jakirkham: Maybe some notification should be sent out, as well? I'm open to ideas to improve it. Though blocking deployment due to some failed feedstock (particularly of orthogonal feedstocks), feels wrong to me.

ocefpaf: > Maybe some notification should be sent out

Yeah. That is better!

jakirkham: Definitely, we don't want quiet failures. I absolutely agree. :+1:

Just at the moment, I'm a little unsure of how to do this. So, this is what I have to fix the current problem.

jakirkham:
While this looks like it should work, I'm going to wait for CI before merging as I don't want to make the existing situation worse.

jakirkham:
Turns out this didn't fix it completely because of GitPython's behavior. See this PR ( https://github.com/conda-forge/conda-forge.github.io/pull/68 ) for the complete fix.

PR #68, opened 2016-03-30
feedstocks: Add submodules quietly

jakirkham:
Fixes: https://github.com/conda-forge/conda-forge.github.io/issues/54
Related: https://github.com/conda-forge/conda-forge.github.io/pull/66

It turns out that GitPython will leave the repo in an incomplete state (modifies `.gitmodules` and keeps the repo around) after it fails to add a submodule. Here we have opted to call `git` directly via GitPython's interface to work around this problem. This allows us to handle the same error as before and not have to worry that the repo is in an incomplete state. For good measure, we clean up the repo that was cloned as `git` does not do this for us. However, it does not leave `.gitmodules` with modifications when it fails.

Tried this out locally to make sure it actually worked as intended.

### DIFF
--- a/scripts/update_feedstocks_submodules.py
+++ b/scripts/update_feedstocks_submodules.py
@@ -10,9 +10,11 @@
 # run_with: python
 
 import os
+import shutil
 
 import conda_smithy.feedstocks as feedstocks
 from git import Repo
+from git.exc import GitCommandError
 
 import argparse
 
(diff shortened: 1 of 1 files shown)


jakirkham:
This works nicely. :smile:

See this [log]( https://travis-ci.org/conda-forge/conda-forge.github.io/jobs/119419944#L228 ) where a warning is issued for an empty repo and then work continues.

PR #69, opened 2016-03-30
Require `git` for steps that use it

jakirkham:
On old systems it is possible to have a version of `git` that does not support all of the arguments used in these scripts. Fortunately, `git` is packaged at conda-forge. :smile: So, we should simply use it when we need it instead of relying on a system to have an acceptable version.


jakirkham:
Will try again once we fix git.

PR #76, opened 2016-04-02
Fix trailing whitespace

jakirkham:
Removes trailing whitespace from `feedstock.html` and ` feedstocks.html.tmpl` to make sure it doesn't get added again. Used this `sed` command to clean up `feedstock.html`.

    sed -i.bkup 's/[ ]*$//' feedstocks.html

### DIFF
--- a/feedstocks.html.tmpl
+++ b/feedstocks.html.tmpl
@@ -65,7 +65,7 @@
             <div class="col-lg-8 col-lg-offset-2">
                 <h2 class="text-center">Feedstocks on conda-forge</h2>
                 <ul class="list-group">
-                  {% for feedstock in gh_feedstocks %}
+{% for feedstock in gh_feedstocks %}
                     <li class="list-group-item">
                         <span class="label pull-right">
                             <div href="https://circleci.com/gh/conda-forge/{{feedstock.package_name}}-feedstock" class="ci-badge">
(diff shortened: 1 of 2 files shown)


feedstocks.html.tmpl
```diff
@@ -65,7 +65,7 @@
             <div class="col-lg-8 col-lg-offset-2">
                 <h2 class="text-center">Feedstocks on conda-forge</h2>
                 <ul class="list-group">
-                  {% for feedstock in gh_feedstocks %}
+{% for feedstock in gh_feedstocks %}
```
jakirkham: After jinja processes this, it leaves the whitespace from this around. Figure it is simpler to de-dent to avoid this as it will parse it the same.

feedstocks.html.tmpl
```diff
@@ -73,16 +73,16 @@
...
                                 <img src="https://ci.appveyor.com/api/projects/status/github/conda-forge/{{feedstock.package_name}}-feedstock?svg=True">
-                            </div>                            
+                            </div>
                         </span>
                         <a href="https://github.com/conda-forge/{{ feedstock.name }}">
                             {{ feedstock.package_name }}
                         </a>
                     </li>
-                  {% endfor %}
+{% endfor %}
```
jakirkham: Same story as for this jinja code.

jakirkham:
Made two comments where it might be interesting. The rest is vanilla white space trimming.

jakirkham:
This ready for review. :smile:

jakirkham:
Tested this locally and it seems to work fine.

PR #98, opened 2016-04-16
Convert `None` to empty strings when rendering jinja.

jakirkham:
Fixes https://github.com/conda-forge/conda-forge.github.io/issues/93

Alright, this should finally fix it. :weary:

Basically, this let's our silent dictionary (`SilentDict`) behave as expected (returning `None` for undefined keys and allowing deletion of undefined keys). On the jinja side though, it converts all `None`s to `""` so that it behaves as if it wasn't there in the first place, but we can treat it like a `dict` that just always uses `get`.

### DIFF
--- a/scripts/update_teams.py
+++ b/scripts/update_teams.py
@@ -68,6 +68,12 @@ class SilentDict(collections.MutableMapping):
         return len(self.store)
 
 
+def jinja_finalize(value):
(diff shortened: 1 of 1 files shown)


jakirkham:
Third time is a charm? :four_leaf_clover:

PR #146, opened 2016-05-18
MNT: Print out which recipe is being checked

ericdill:
This will help figure out exactly what recipe is raising an exception during
parsing.  Right now we simply have no idea if travis fails.


ericdill:
@jakirkham Can you merge this so we can see which recipe is throwing this error? https://travis-ci.org/conda-forge/conda-forge.github.io/jobs/131008956#L244

jakirkham:
Absolutely! It's in.

PR #148, opened 2016-05-20
Require `git` for steps that use it (Pt. 2)

jakirkham:
As our `git` now works correctly with certs, we are re-adding `git` to the requirements of scripts that use `git`.

xref: https://github.com/conda-forge/conda-forge.github.io/pull/69
xref: https://github.com/conda-forge/conda-forge.github.io/pull/71

### DIFF
--- a/scripts/lint_feedstocks.py
+++ b/scripts/lint_feedstocks.py
@@ -2,6 +2,7 @@
 
 # conda execute
 # env:
+#  - git
 #  - python
 #  - conda-smithy
 #  - gitpython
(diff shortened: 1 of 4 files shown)


ericdill:
Why is gitpython explicitly listed as a dependency?

jakirkham:
Because we import it [here]( https://github.com/conda-forge/conda-forge.github.io/pull/148/files#diff-4d91de38bfb71e44cf367f1441144fdfR27 ). I think we have gotten away with this somehow probably because something else is pulling in the dependency and thus it is already there.

ericdill:
Oh right, sorry.  I was looking for `import gitpython`.

jakirkham:
No worries. I had to double-check myself as I don't play with this code very often.

jakirkham:
Alright, in it goes.

PR #182, opened 2016-06-30
Update xz and sqlite

ocefpaf:
Update to `xz 5.2.*` and to `sqlite 3.13.*`


jakirkham:
Can we getting pinning PR updates for this? The stack is in-between two pinnings and it is causing us issues.

Issue #75, opened 2016-04-02
selecting text on chrome


janschulz:
When I select text on the website (e.g. the config addition), there is no selection marker (-> no blue background), but it seems the text is selected and is copy-abel (sp?).

It would be nice to get a visual response when I want to copy something.

jcb91:
Although the non-styled selection might be useful/necessary for some things, I think it's a reasonable assumption that anything in a `code` tag should probably not be subject to this...

jakirkham:
Believe this is fixed by PR ( https://github.com/conda-forge/conda-forge.github.io/pull/186 ). Reopen or re-raise if this is not the case.

PR #186, opened 2016-07-04
allow selection highlighting on code tags

jcb91:
as mentioned in #75

### DIFF
--- a/css/theme.css
+++ b/css/theme.css
@@ -379,8 +379,13 @@ footer p {
     margin: 0;
 }
 
-::-moz-selection:not(code),
-::selection:not(code) {
+:not(code)::-moz-selection {
+    text-shadow: none;
+    background: #fcfcfc;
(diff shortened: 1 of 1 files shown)


jakirkham:
So, I'm not quite sure what the original request was, but this has an undesirable effect IMHO. Basically selection highlighting doesn't seem to work any more. For instance, if I click on "About" or "Contribute" or scroll to those sections, the text in the title is no longer highlighted. See before...

screen shot 2016-07-05 at 0 00 35

...and after...

screen shot 2016-07-04 at 23 58 46

jcb91:
oh wow, strange unintended consequences! That doesn't happen for me, and, to be honest, I can't see why that would happen, since the altered rules don't affect those elements anyway...

jakirkham:
Yeah, I still got that behavior, but I think it is htmlpreview being weird. Tried with rawgit and it worked fine.

Thanks @jcb91.

PR #198, opened 2016-07-27
Pin OpenBLAS to 0.2.18

jakirkham:
Appears we are not pinning OpenBLAS here. So this changes that. For the most part, this is the pinning we are using throughout conda-forge. However, there are a few notable exceptions that also need to be updated to follow the BLAS standard.

cc @pelson @conda-forge/openblas


jakirkham:
Thoughts on this?

patricksnape:
All good from me.

PR #199, opened 2016-07-30
Pin geos

ocefpaf:
As `defaults` just updated their `geos` I realized we should be pinning it here to avoid conflicts.


jjhelmus:
LGTM

PR #206, opened 2016-08-12
Add general meeting minutes for 2016-07-22

jakirkham:
This is a markdown converted copy of the meeting minutes from the general meeting on 2016-07-22 for archival. Please let me know if there need to be any changes. Will merge in 24hrs if there are no changes.

In attendance: @jakirkham @jjhelmus @mwcraig @pelson @msarahan @ocefpaf

### DIFF
--- a/minutes/2016-07-22.md
+++ b/minutes/2016-07-22.md
@@ -1,0 +1,231 @@
+## 2016-07-22: General discussion
+
+Time: 14:00 UTC
+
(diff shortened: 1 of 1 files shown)


mwcraig:
LGTM, but it would be nice to indicate in some way which topics were actually discussed and which we didn't get to. 

Maybe a section heading like `Deferred items` before the chunk of ones that were not discussed.

jakirkham:
This isn't a bad idea.

We also talked briefly about not copying the agenda from meeting to meeting to encourage people to raise important issues again on their own. Though it is has its own pros and cons.

Oddly enough we probably should make this a discussion item next meeting.

In any event, I don't think the time to change the format is when we archive. It should happen as part of the meeting in the Hackpad before we archive.

mwcraig:
That makes sense.

jakirkham:
Thanks @mwcraig.

Went ahead and put this in as there were no other things to address. Plus we probably should archive it before the next meeting. 😉 If there are some pressing changes needed, please raise them in a subsequent PR.

PR #212, opened 2016-08-19
Drops CI badges from feedstocks page

jakirkham:
Drops the CI badges from feedstocks page. This is done as loading all of these badges is quite expensive for the end user, GitHub, and the CIs. If we want to capture this statistic, we should find a new way of doing it as this is no longer performant and causes trouble for others.

cc @pelson


jakirkham:
Was there an issue for removing the badges? I thought there was, but I couldn't find it.

PR #215, opened 2016-08-20
Use generic VM for Travis CI

jakirkham:
Switch to using the generic VM for Travis CI as we use conda to get everything we need. This should improve VM boot time.


jakirkham:
Going to take this change for a spin. This should be trivial to revert if it does run into issues. However, there really should be any as we use Miniconda's Python not Travis' Python.

jakirkham:
Works fine.

Issue #150, opened 2016-05-20
Generating our own ABI reports


jakirkham:
It has become such a common thing to worry about ABI compatibility here that I wonder if we shouldn't have a mechanism to generate and publish our own ABI reports using binaries we have built. This probably will be really useful in a few months. Most of the time spent in the near term will just be trying to determine and summarize this information.

janschulz:
See also https://www.debian.org/doc/debian-policy/ch-sharedlibs.html#s-sharedlibs-symbols which is probably exactly what you are asking for here...

jakirkham:
I'm thinking of something mind numbingly simple. For instance, this ABI report for [FFmpeg]( http://abi-laboratory.pro/tracker/timeline/ffmpeg/ ). There are a bunch of Perl [tools]( http://abi-laboratory.pro/open-source.html ) that we could use to do this or we can take inspiration from them and add similar things to our own Python scripts or `conda` directly.

janschulz:
You need a DB for that, to collect the list of Symbols for each version or at least the last one... The debian script seems to do something similar by just looking at a file in the package -> if the file has something different, then it asks you to update the file which in turn triggers the increased versions for the dependency lines in the packages which depend on that package.

Windows could be a problem, because I guess that the ABI could be different for *ix and win.

jakirkham:
xref: https://github.com/conda-forge/staged-recipes/issues/1601

cc @gqmelo

jakirkham:
cc @patricksnape @183amir

patricksnape:
I'm leaning towards thinking that this would be very useful.

PR #233, opened 2016-09-07
Pin `libevent` to `2.0.*`

jakirkham:
Trying something different by proposing the pinning before we get a bunch of stuff built without a pinning. 😉

Based off of the linked ABI table. There seem to be a few big changes in the early part of the 2.0.x series. However, we are starting with 2.0.22 so am hoping that we will avoid further breaks by not going backwards.

http://upstream.rosalinux.ru/versions/libevent.html

xref: https://github.com/conda-forge/staged-recipes/pull/1491
xref: https://github.com/conda-forge/staged-recipes/pull/1507

### DIFF
--- a/scripts/pin_the_slow_way.py
+++ b/scripts/pin_the_slow_way.py
@@ -36,6 +36,7 @@ pinned = {
           'icu': 'icu 56.*',
           'jpeg': 'jpeg 9*',
           'libblitz': 'libblitz 0.10|0.10.*',
+          'libevent': 'libevent 2.0.*',
           'libmatio': 'libmatio 1.5.*',
           'libnetcdf': 'libnetcdf 4.4.*',
           'libpng': 'libpng >=1.6.21,<1.7',


jakirkham:
Thoughts on this, @conda-forge/core?

patricksnape:
Fine with me. As a note, those ABI trackers look super useful, but that one is no longer maintained. I saw that @183amir used [this one](http://abi-laboratory.pro/tracker/) for the GIFlib pinning - which is actually [open source](https://github.com/lvc/abi-tracker) - perhaps we could see about adding more libraries that we are pinning to http://abi-laboratory.pro or setting up our own ABI tracker? We might even be able to set it up as an optional part of the conda builds?

183amir:
lvc said in comments a few days ago:

> Feel free to add more libraries to the tracker (add an issue to https://github.com/lvc/upstream-tracker/issues). I'm interested to add all "live" and popular libraries.
Thank you.

jakirkham:
Cool, asked if they can added it in issue ( https://github.com/lvc/upstream-tracker/issues/2 ).

jakirkham:
Also issue ( https://github.com/conda-forge/conda-forge.github.io/issues/150 ) may be of interest to both of you.

scripts/pin_the_slow_way.py
```diff
@@ -36,6 +36,7 @@
           'icu': 'icu 56.*',
           'jpeg': 'jpeg 9*',
           'libblitz': 'libblitz 0.10|0.10.*',
+          'libevent': 'libevent 2.0.*',
```
183amir: Based on http://abi-laboratory.pro/tracker/timeline/libevent/ you might want to pin like this: `libevent >=2.0.19,<2.1.*`

jakirkham: So we only have `libevent` version `2.0.22`. `defaults` only has `libevent` versions `2.0.20` and `2.0.21`. As such, I don't think that lower bound is necessary and this should be, for all practical concerns, fine.

183amir: I would say this is more about good practices in our pinnings rather than what can happen in reality but anyways I think this pull that I opened needs some love: https://github.com/conda-forge/conda-forge.github.io/pull/245

jakirkham: It is a fair point. However, `libevent` is only used in one dependency that is already pinning it this way.

I would love to see some more discussion on your PR, but expect it will happen after switching to `conda-build` version 2.x.

jakirkham:
Given this pinning is already in use, going to merge this as is. Though I agree with @183amir that we should be taking a closer looking at PR ( https://github.com/conda-forge/conda-forge.github.io/pull/245 ).

PR #235, opened 2016-09-08
Pin giflib

183amir:
`giflib` is known to break backward compatibility sometimes on a patch version
 in the past: http://upstream.rosalinux.ru/versions/giflib.html
But new results: http://abi-laboratory.pro/tracker/timeline/giflib/ show that it is following semantic versioning.

### DIFF
--- a/scripts/pin_the_slow_way.py
+++ b/scripts/pin_the_slow_way.py
@@ -32,7 +32,7 @@ pinned = {
           'fontconfig': 'fontconfig 2.11.*',
           'freetype': 'freetype 2.6.*',
           'geos': 'geos 3.4.*',
-          'giflib': 'giflib 5.1.2|5.1.2.*',
+          'giflib': 'giflib 5.1.*',
           'hdf5': 'hdf5 1.8.17|1.8.17.*',
           'icu': 'icu 56.*',
           'jpeg': 'jpeg 9*',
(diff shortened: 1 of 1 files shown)


scripts/pin_the_slow_way.py
```diff
@@ -32,6 +32,7 @@
           'fontconfig': 'fontconfig 2.11.*',
           'freetype': 'freetype 2.6.*',
           'geos': 'geos 3.4.*',
+          'giflib': 'giflib 5.1.2|5.1.2.*',
```
patricksnape: According to that site they've never released a ``5.1.2.*`` type version? Can we guarantee if they did it wouldn't be breaking? Or should we just hard pin to `5.1.2`?

183amir: My recommendation is a hard pin on 5.1.2 but people here are usually against it.

183amir:
Given the new backward compatibility results here: http://abi-laboratory.pro/tracker/timeline/giflib/
I have patched `giflib` on a minor version.

Please merge this at your earliest convenience.

183amir:
Thanks :+1:

jakirkham:
Appears they added a new symbol in 5.1.2. Not sure which side of the breaking we want to be on, but would pick one.

PR #247, opened 2016-10-04
Tweaks to fork feedstocks

jakirkham:
This is a follow-up on PR ( https://github.com/conda-forge/conda-forge.github.io/pull/242 ). Still forks feedstocks one maintains as before, but makes the following tweaks.

* Picks a more sensible default argument.
* Adds the remotes by user/org locally.
* If remotes already exist, skip them.


jakirkham:
Toggling to try to get CI to start. Not that it is particularly relevant.

PR #248, opened 2016-10-05
Pin qt/pyqt

ocefpaf:
Now that `defaults` moved to Qt 5 we should pin our packages to `pyqt 4.11.*` and `qt 4.8.*` until we have versions of `Qt 5` and `pyqt 5` compiled with `jpeg 9*`.


ocefpaf:
Pinning `matplotlib` as suggested here fixed conda-forge/modflow2netcdf-feedstock#3, conda-forge/windrose-feedstock#3, and conda-forge/flopy-feedstock#6

conda-forge/mplleaflet-feedstock#3 is pending due to the fact that conda insists on getting some packages from `defaults`. I am debugging that.

In light of that I am self-merging this. Hopefully we will be able to move to `Qt`/`pyqt 5` soon.

PR #255, opened 2016-10-18
Fix dependency matching

ocefpaf:
Should help with https://github.com/conda-forge/instaseis-feedstock/pull/5/files#r83806867

Ping @krischer. I am not sure we need we need `regex` but if you want to give it a try I can close this one. (Not sure this will work... It is too early in the morning here and I should not be sending PRs :smile:)

### DIFF
--- a/scripts/pin_the_slow_way.py
+++ b/scripts/pin_the_slow_way.py
@@ -276,7 +276,7 @@ for feedstock, git_ref, meta_content, recipe in feedstock_gen:
 
                 for pos, dep in enumerate(section):
                     for name, pin in pinned.items():
-                        if dep.startswith(name) and dep != pin:
+                        if dep.split()[0] == name and dep != pin:
                             replacements['- ' + str(dep)] = '- ' + pin
             if replacements:
                 current_build_number = recipe['build']['number']


krischer:
This, if I read the code correctly, only fixes part of the problem. The `meta.yaml` in the instaseis receipe contains

```yaml
 - pyqt
 - pyqtgraph
```

The current logic will create a replacement dictionary akin to

```python
replacements = {
    "- pyqt": "- pyqt 4.11.*"
}
```

This will also be applied to the `pyqtgraph` item in the `meta.yaml`...

Additionally it will also trigger for receipts that already pin their pyqt version which is likely undesired. I don't think there is a simpler way than regular expressions for safe behaviour. They are needed to create the replacement dictionary but also to apply it.

ocefpaf:
> This will also be applied to the `pyqtgraph` item in the `meta.yaml`...

Not sure that is true. I am using `dep.split()[0] == name` which should match only `pyqt` and not `pyqtgraph`.

> Additionally it will also trigger for receipts that already pin their pyqt version which is likely undesired. I don't think there is a simpler way than regular expressions for safe behaviour. They are needed to create the replacement dictionary but also to apply it.

It is untested but I also believe that is not true. If `dep` is already `pyqt 4.11.*` than it is equal to `pin` and will pass. No? Maybe I am misinterpreting that is passed as `dep` here...

Ping @pelson.

PS: I have a CLI pin version of this and I tested there, but my version diverged quite a bit from the original one. Maybe we should re-factor this version into a CLI that is used by the webservice to avoid the forking. (I believe a CLI is quite useful for re-builds, linting, or just pining locally before pushing to `conda-forge`.)

krischer:
> This will also be applied to the pyqtgraph item in the meta.yaml...
> 
> Not sure that is true. I am using dep.split()[0] == name which should match only pyqt and not pyqtgraph.

Not in this line but the later logic that actually replaces the string. That's what went wrong in the instaseis example - it never actually got to replace the `pyqtgraph` line because it was already changed.



> It is untested but I also believe that is not true. If dep is already pyqt 4.11.* than it is equal to pin and will pass. No? Maybe I am misinterpreting that is passed as dep here...

True - but it might well be that someone pins to some other version in which case it would replace it.

In any case #256 should solve all these problems.

ocefpaf:
> True - but it might well be that someone pins to some other version in which case it would replace it.

Not that may be desirable though. The pinning script does not know that version the maintainer wants and suggests one that is consistent with the rest of the stack in `conda-forge`. It is up to the maintainer to accept or close the PR. So the script should always suggest the pinning here.

> Not in this line but the later logic that actually replaces the string. That's what went wrong in the instaseis example - it never actually got to replace the `pyqtgraph` line because it was already changed.

I will look into that. It does not happen in my CLI version of this script though.

ocefpaf:
@pelson please look #256 before considering this. It is probably a better implementation. My changes here are based on a slightly different version of `pin_the_slow_way.py`. See https://github.com/conda-forge/conda-forge.github.io/pull/255#issuecomment-254464626

ocefpaf:
Putting my hopes on #256 and making this easier for @pelson to review :stuck_out_tongue_winking_eye:

PR #260, opened 2016-11-02
Pin libxml2

ocefpaf:
It seems that `libxml2 2.9.4` has a bug so it is safer to pin to `2.9.4` until we sort this out. See https://mailman-mail5.webfaction.com/pipermail/lxml/2016-September/007759.html

Ping @rsignell-usgs and @kwilcox 

PS: I am pining `lxml` and `pycsw` in a moment.

### DIFF
--- a/scripts/pin_the_slow_way.py
+++ b/scripts/pin_the_slow_way.py
@@ -42,6 +42,7 @@ pinned = {
           'libpng': 'libpng >=1.6.23,<1.7',
           'libsvm': 'libsvm 3.21|3.21.*',
           'libtiff': 'libtiff 4.0.*',
+          'libxml2': 'libxml2 2.9.3',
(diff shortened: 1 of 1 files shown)


jschueller:
I guess it's this one: https://bugzilla.gnome.org/show_bug.cgi?id=766834 ?
(it shows up in the important debian bugs: https://bugs.debian.org/cgi-bin/pkgreport.cgi?package=libxml2, plus it's the one in 2.9.4)
It was fixed upstream, I suggest we backport it instead of pinning.

ocefpaf:
> I suggest we backport it instead of pinning.

That is a nicer approach! Feel free to send a PR to the `libxml2` feedstock.

jschueller:
already on it!

jschueller:
https://github.com/conda-forge/libxml2-feedstock/pull/23

PR #261, opened 2016-11-06
Update boost pinning to 1.62

jschueller:


jschueller:
@ocefpaf this one too

PR #262, opened 2016-11-07
Update openblas pinning to 0.2.19

jschueller:


scopatz:
I think that this can be merged.

PR #265, opened 2016-11-10
try to build and deploy docs to same domain

scopatz:
Rather than just to RTD


scopatz:
@conda-forge/core, can someone take a look at this please? Thanks!

pelson:
I'm in favour of this change. @jakirkham, as you setup RTD, are you happy to merge?

jakirkham:
Anyone who wants to carry forward on documentation gets no blockers from me. 😄

Mainly setting up RTD was less work for me. This is probably the nicer way to do things.

Two questions. Have we tested this locally (this isn't covered by CI because of the `GH_TOKEN`)? Also can we please have a link from the homepage to the docs? Once answered/addressed, I think we should be good to go.

scopatz:
> Have we tested this locally (this isn't covered by CI because of the GH_TOKEN)? 

This is hard to test outside of the context of CI, but it does seem to run for me if I comment out most of the CI stuff.

> Also can we please have a link from the homepage to the docs? 

Yes, but I wanted to make sure that this builds and uploads on CI first. I was going to add that in another PR. No sense in having a broken link on the website.

jakirkham:
Ok, let's try then.

jakirkham:
Also would like to point out that we should get moving on breaking out team/feedstocks update scripts to a webservice (issue xref'd). Making lots of docs updates is going to make us hit our rate limit real fast blocking further doc updates for awhile.

I'd even go so far to say that the feedstock listing could even be broken out of the webpage repo proper and moved to the feedstocks repo too. We can even host it under ( https://conda-forge.github.io/feedstocks ), which is already prepped to do exactly this. 

xref: https://github.com/conda-forge/conda-forge-webservices/issues/63

.ci_scripts/update_docs
```diff
@@ -12,6 +11,11 @@ if [ -n "$GH_TOKEN" ]; then
...
     # Generate html requires a GitHub token in order to query the available feedstocks.
     conda execute scripts/generate_html.py
+    conda install sphinx --yes --quiet -c conda-forge
+    cd src
+    make html
+    cp -r _build/html/* ../docs
+    rm -rf _build
     git diff
     # Only build the docs for commits on master.
     if (( "$TRAVIS_BRANCH" == "master" && "$TRAVIS_PULL_REQUEST" == "false" )); then
```
jakirkham: So this didn't work. It is because we are using `git commit -a` instead of `git add -A` and then `git commit`. The former only stages files that were modified, but not new files. The latter adds new files. Could you please add a follow-up PR to change this, @scopatz ?

jakirkham: Addressing in PR ( https://github.com/conda-forge/conda-forge.github.io/pull/271 ).

jakirkham: More tweaks required. Done in PR ( https://github.com/conda-forge/conda-forge.github.io/pull/272 ).

jakirkham:
Had to make some tweaks to get this to work, but it now works.

screen shot 2016-11-25 at 20 12 57

jakirkham:
Adding a simple link to the docs from the about section in PR ( https://github.com/conda-forge/conda-forge.github.io/pull/273 ). Please take a look.

PR #279, opened 2016-11-29
add note on repo tarballs

scopatz:
This is a docs change.

### DIFF
--- a/src/meta.rst
+++ b/src/meta.rst
@@ -1,0 +1,22 @@
+Writing the ``meta.yaml``
+==========================
+This document presents conda-forge rules, guidelines, and justifications
+about writing the ``meta.yaml`` file.
+
+
+Build from Tarballs, Not Repos
+------------------------------
+Conda-forge requires that building from tarballs using the
+``url`` and ``fn`` keys in the ``build`` section. A recipe
+should not use the ``git_url``, ``git_ver``, and similar
+keys. There are two main reasons for this:
+
+* Downloading the whole repo when you only need a single snapshot wastes
+  the precious, constrained, and shared CI time and bandwidth.
+* Repositories are not checksummed. Thus, using a tarball has a
+  stronger guarantee that the download that is obtained to build from is
+  in fact the intended package.
+
+If a package does not have the ability to build from a tarball, this is
+considered a bug and should be reported upstream. In the worst case,
+the source can be patched in include the relevant build information.
--- a/src/index.rst
+++ b/src/index.rst
@@ -12,6 +12,7 @@ Contents:
    :maxdepth: 2
 
    guidelines
+   meta
 
 
 Indices and tables


scopatz:
Yeah, I agree that there may be some extenuating circumstances in practice. I was trying to write this in such a way that people submitting everyday packages, though, don't think that they are somehow the exception. All other options should typically be exhausted first.

src/meta.rst
```diff
@@ -0,0 +1,24 @@
...
+  the precious, constrained, and shared CI time and bandwidth.
+* Repositories are not checksummed. Thus, using a tarball has a
+  stronger guarantee that the download that is obtained to build from is
+  in fact the intended package.
+* On some systems (such as Windows), it is possible to not have permissions
+  to remove a repo once it is created. This can be avoided by using a tarball.
+
+If a package does not have the ability to build from a tarball, this is
+considered a bug and should be reported upstream. In the worst case,
+the source can be patched in include the relevant build information.
```
shadowwalkersb: Typo? Maybe, "the source can be patched in to include the relevant build information."?

shadowwalkersb:
Is there an easy way to build a tarball for git projects that have submodules? GitHub does seems to provide only the projects code in the tarballs, but not submodules. Maybe, it can be mentioned as an exception? Calling it a bug is a bit harsh, I think.

src/meta.rst
scopatz: thanks!

scopatz:
My (imperfect) understanding of how most sub-modules are used is to acquire dependencies needed for the build, if they are not otherwise available. In this case, the dependencies should get their own packages prior to building the package at hand.

Alternatively, if the sub-modules build core behaviour for the package, the build system should provide a way for getting the code. If it doesn't, I do consider that a deficiency in the build system. Furthermore, in the worst case scenario, the recipe build script itself can always grab and unpack the tarballs for all of the sub-modules itself. 

I think it is always possible to avoid going from the repo in the face of submodules.

minrk:
> Is there an easy way to build a tarball for git projects that have submodules?

Yes, there's [git-archive-all](https://github.com/Kentzo/git-archive-all) which does what (I think) git-archive should always do. But it needs to be run somewhere in order to do the bundling.


I think there's a distinction between "from a tarball" and "from the git-archive result you get from GitHub's auto-generated tarballs". For a while, IPython tracked js dependencies in a separate repo pulled in from a submodule. IPython sdists included the contents of the submodule, but git-archive excludes submodules. This meant that you *could not* install IPython from the GitHub tarballs, you had to use a real git checkout or the proper sdists on PyPI. I (still) view this as a bug in git archive, rather than IPython's build system. However, given this deficiency, a project that relies on submodules should not rely on git-archive-generated tarballs, and should produce tarballs that actually work (as IPython did with sdists). Similarly, I would recommend that all conda-forge packages avoid using the GitHub generated tarballs unless there is no other option (e.g. PyPI). Installing from a raw git repo without git info is not generally safe or supported. Plus, I don't think GitHub makes any guarantee of the checksum-stability of its generated files.

jakirkham:
Sadly there are lots of things about git submodules that are half-baked. `git archive` is only one example. Here's an [article (rant)]( https://codingkilledthecat.wordpress.com/2012/04/28/why-your-company-shouldnt-use-git-submodules/ ) on some of the pain points and a [more balanced one]( http://blogs.atlassian.com/2013/03/git-submodules-workflows-tips/ ). That being said, if someone knows what they are doing, git submodules can be used effectively or they can use [something better]( https://blogs.atlassian.com/2013/05/alternatives-to-git-submodule-git-subtree/ ). Personally I understand the use cases for submodules and use them occasionally when needed. That being said, there are known issues and one needs to be prepared for them and address them somehow.

> Plus, I don't think GitHub makes any guarantee of the checksum-stability of its generated files.

So this is sort of complicated. I had a discussion with GitHub about this. They generate all of their archives on-demand (likely using `git archive` and some other stuff though they didn't say that). However, they try to run a number of checks when they change the tarball generation process to make sure the checksum does not change.

However, that process does occasionally fail as happened to us in September. Here is [one case]( https://github.com/conda-forge/staged-recipes/pull/1562 ) where it failed, but it actually affected all GitHub autogenerated archives. So many people had to change their checksum for the same release. GitHub recommended to me that if we care about the checksum being the same for a tarball, we should be publishing our own tarballs as part of the release and using them.

TBH I agree with @minrk and GitHub that this is really what should be done by developers on GitHub. Though we (conda-forge members) don't often have control over this (unless we are the developers too). Alternatively going to PyPI provides guarantees w.r.t. checksums when it comes to Python packages. Google does a good job of publishing their tarballs. If there is nothing better than a GitHub generated archive, I understand we will sometimes need to use them. However, maintainers need to be aware of the aforementioned problems and be able to address them when they come up.

scopatz:
Merging, please feel free to PR in any further changes and I will be happy to review.

PR #285, opened 2016-12-13
pinned pango to 1.40.*

croth1:
We decided to pin pango here: https://github.com/conda-forge/staged-recipes/pull/2066.


ocefpaf:
Thanks @croth1! 

@pelson I am guessing the `in progress` label is some sort of automate tool you are using and no really mean this is in progress, right?

PR #299, opened 2016-12-28
added Circle docs

scopatz:


scopatz:
Yeah, if the issue ever does go away, I'd be happy to remove these docs ;)

jakirkham:
> Yeah, if the issue ever does go away, I'd be happy to remove these docs ;)

Could I take you on this offer now that things seem to be working more smoothly? 😉

That being said, I do think it is handy to remind people how to setup builds on their own forks. Sometimes it takes a while to see your build run at conda-forge due to the demand. Also experimentation with feedstocks that have long build times are often better dealt with on a fork initially. Maybe it would be handy to keep this, but adjust a few things. We could also add notes for the other CI services.

scopatz:
Well, it still isn't getting started up for all of the feedstocks, right? And there is no public API to enable it, correct? Maybe a rephrasing is needed, but I not sure we should remove it until it works by default everywhere.

jakirkham:
> Well, it still isn't getting started up for all of the feedstocks, right?

Actually it is. We had been using this on nearly all feedstocks except brand new ones for the past few days. I ran it again to enable it on all recently added feedstocks (past couple days) and everything else for good measure. Also new feedstocks will get this enabled out of the box thanks to the new `conda-smithy`. Will of course be watching this closely in the near term, but am confident from our testing and recent usage that this works.

> And there is no public API to enable it, correct?

There is no public API in the sense that CircleCI is hesitant to document and have the masses start to use it. TBH that is totally reasonable given it is a new feature that they added and they/we are trying it out. I'm not sure it would be fair for us to require they make it official in such a short time frame. That being said, I emailed them about our use case and told them what we are doing. They responded saying that it was ok for us to be using the API in this way.

> Maybe a rephrasing is needed, but I not sure we should remove it until it works by default everywhere.

Agreed, which was what I was getting to at the end of that comment. Whether this works or not, it is always handy to know how to enable CIs on one's forks for a variety of reasons. So I'm ok with this being there. That being said, we should probably encourage trying without it and/or enabling this through the CircleCI settings GUI if needed (as a few non-admin users have been able to do). There are some benefits to testing forked PR builds instead of testing on one's fork (e.g. testing the merge commit).

scopatz:
Does #312 satisfy this for now? I am sorry I didn't really understand your second paragraph originally.

PR #300, opened 2016-12-28
added windows build instructions

scopatz:
I always forget how to do this, so I thought this would be a helpful reference.

### DIFF
--- a/src/buildwin.rst
+++ b/src/buildwin.rst
@@ -13,6 +13,8 @@ build for such projects.
 
 .. code-block:: bat
 
+    setlocal EnableDelayedExpansion
+
     :: Make a build folder and change to it.
     mkdir build
     cd build
--- a/src/buildwin.rst
+++ b/src/buildwin.rst
@@ -1,0 +1,32 @@
+Building on Windows
+==========================
+This document presents conda-forge and conda-build information and examples
+when building on Windows.
+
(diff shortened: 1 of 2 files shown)


src/buildwin.rst
```diff
@@ -0,0 +1,32 @@
...
+------------------------------
+Some projects provide hooks for CMake to build the project. The following
+example ``bld.bat`` file demonstrates how to build a traditional, out-of-core
+build for such projects.
+
+**CMake-based bld.bat:**
+
+.. code-block:: bat
+
+    setlocal EnableDelayedExpansion
```
jakirkham: I don't think we generally need this. @msarahan @mingwandroid , thoughts?

scopatz:
Thanks for the feedback, I have updated the PR to reflect the comments.

src/buildwin.rst
msarahan: Better to keep it, IMHO. It leads to more intuitive behavior.

scopatz: Added it back, and I agree.

jakirkham: Ok. Should we add `@echo off` too?

scopatz: I don't think it matters much, but my inclination is to have echo on, like with bash and `set -x`

PR #318, opened 2017-01-20
Update feedstock submodules manually

jakirkham:
Instead of running several implicit `for`-loops over the submodules in the `feedstocks` repo, run one update loop manually that fits the same criteria. Also print out info about each feedstock that is being updated. This last part is particularly important for Travis CI, which expects some kind of output every 10mins. This should easily meet that criteria.


jakirkham:
This should fix generation of the `feedstocks` repo so that it works again.

PR #320, opened 2017-01-21
Temporarily use Python 2 for team updates

jakirkham:
Ran into what appears to be a bug with `conda-build-all` and Python 3 during team updates. It appears the problem goes away with Python 2. An issue has been filed upstream about the problem. In the interim as this does workaround the issue, would like to switch to Python 2 to clear the current batch of recipes. May be worthwhile keeping this switch until the cause of the bug is better understood and potentially fixed or worked around in a better way.

xref: https://github.com/SciTools/conda-build-all/issues/77

### DIFF
--- a/scripts/update_teams.py
+++ b/scripts/update_teams.py
@@ -2,7 +2,7 @@
 
 # conda execute
 # env:
-#  - python
+#  - python 2.7.*
 #  - conda-smithy
 #  - pygithub 1.*
 #  - six


jakirkham:
Help figuring out and/or solving the underlying issue with `conda-build-all` in issue ( https://github.com/SciTools/conda-build-all/issues/77 ) appreciated. In the interim, just trying to get the team update script running again after a 5 day hiatus.

Issue #234, opened 2016-09-07
rpy2 installation linking problems


jbwhit:
I was having problems that I couldn't diagnose, so I deleted all of miniconda/.jupyter/etc, and reinstalled.
## The error

``` python
import rpy2
%load_ext rpy2.ipython
```

```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-3-a69f80d0128e> in <module>()
----> 1 get_ipython().magic('load_ext rpy2.ipython')

/Users/jonathan/miniconda3/envs/problem/lib/python3.4/site-packages/IPython/core/interactiveshell.py in magic(self, arg_s)
   2161         magic_name, _, magic_arg_s = arg_s.partition(' ')
   2162         magic_name = magic_name.lstrip(prefilter.ESC_MAGIC)
-> 2163         return self.run_line_magic(magic_name, magic_arg_s)
   2164
   2165     #-------------------------------------------------------------------------

/Users/jonathan/miniconda3/envs/problem/lib/python3.4/site-packages/IPython/core/interactiveshell.py in run_line_magic(self, magic_name, line)
   2082                 kwargs['local_ns'] = sys._getframe(stack_depth).f_locals
   2083             with self.builtin_trap:
-> 2084                 result = fn(*args,**kwargs)
   2085             return result
   2086

<decorator-gen-64> in load_ext(self, module_str)

/Users/jonathan/miniconda3/envs/problem/lib/python3.4/site-packages/IPython/core/magic.py in <lambda>(f, *a, **k)
    191     # but it's overkill for just that one bit of state.
    192     def magic_deco(arg):
--> 193         call = lambda f, *a, **k: f(*a, **k)
    194
    195         if callable(arg):

/Users/jonathan/miniconda3/envs/problem/lib/python3.4/site-packages/IPython/core/magics/extension.py in load_ext(self, module_str)
     64         if not module_str:
     65             raise UsageError('Missing module name.')
---> 66         res = self.shell.extension_manager.load_extension(module_str)
     67
     68         if res == 'already loaded':

/Users/jonathan/miniconda3/envs/problem/lib/python3.4/site-packages/IPython/core/extensions.py in load_extension(self, module_str)
     82             if module_str not in sys.modules:
     83                 with prepended_to_syspath(self.ipython_extension_dir):
---> 84                     __import__(module_str)
     85             mod = sys.modules[module_str]
     86             if self._call_load_ipython_extension(mod):

[...]

PR #325, opened 2017-01-30
Notes on conda-forge and defaults incompatibilities

ocefpaf:
Based on @mingwandroid's https://github.com/conda-forge/conda-forge.github.io/issues/234#issuecomment-275955636 I decided to start this doc page. It is not ready yet and I appreciate some feedback. The goal is to close any issue related to packages incompatibilities and guide users to this page. (I usually answer 3-5 questions like that a week!)


src/defaults_vs_conda-forge.rst
```diff
@@ -0,0 +1,47 @@
...
+
+How to fix it?
+--------------
+
+Newer `conda` versions introduced a channel priority feature.
+See https://conda.io/docs/channels.html for more information.
+
+One possible solution is to add the `conda-forge` channel on top of `defaults` in your `condarc` file when using `conda-forge` packages.
+This will ensuring that all the dependencies will come from the `conda-forge` channel.
+Here is how a `.condarc` file would look like:
```
ChrisBarker-NOAA: add here:

command:
.. code-block:: shell

    conda config --add channels conda-forge

should do it. after this command, your `.condarc` file should look like:

src/defaults_vs_conda-forge.rst
```diff
@@ -0,0 +1,47 @@
...
+
+.. code-block:: shell
+
+    $ cat .condarc
+    channels:
+      - conda-forge
+      - defaults
+
```
ChrisBarker-NOAA: You can also edit the `.condarc` file directly.

src/defaults_vs_conda-forge.rst
```diff
@@ -0,0 +1,47 @@
...
+In addition to the channel priority we recommend to always install your packages inside a new environment instead the root environment from anaconda/miniconda.
+Using envs make it easier to debug problems with packages and ensure the stability of your root env.
```
ChrisBarker-NOAA: However even in a separate environment, sometimes libraries can be picked up from the root environment, if, for instance a library package is not installed in the current environment at all. If you are getting errors that show a libary mismatch, and the "wrong" version is in the root environment, it may mean you need to install the "correct" version in the environment.

src/defaults_vs_conda-forge.rst
```diff
@@ -0,0 +1,47 @@
...
+It is quite common to install a package from conda-forge and,
+when trying to use it,
+see an error like (OS X example):
+
+.. code-block:: shell
+
+    ImportError: dlopen(/Users/userX/miniconda3/envs/problem/lib/python3.4/site-packages/rpy2/rinterface/_rinterface.so, 2): Library not loaded: @rpath/libicuuc.54.dylib
+      Referenced from: /Users/userX/miniconda3/envs/problem/lib/python3.4/site-packages/rpy2/rinterface/_rinterface.so
+      Reason: image not found
+
```
alimanfoo: Worth to add Linux example as well so people doing a web search for the error message might find their way here too? 

```
.. code-block:: shell

    ImportError: libicuuc.so.54: cannot open shared object file: No such file or directory
```

ResidentMario:
Personally I wonder whether or not this should expanded into a "Gotchas" page. I expect that there are a few issues that are common to builds that would be well addressed by the creation of a canonical page to collect them all.

jakirkham:
From a cursory look, this seems like a nice add. The comments noted above by others seems like reasonable tweaks.

jakirkham:
> Personally I wonder whether or not this should expanded into a "Gotchas" page. I expect that there are a few issues that are common to builds that would be well addressed by the creation of a canonical page to collect them all.

Definitely! Are there any you have experienced that would be worth adding into another PR?

src/defaults_vs_conda-forge.rst
ocefpaf: We should have a "general how to `conda-forge`" I guess. Otherwise we will be asking people to issue that command everywhere. I just finished writing a `conda-smithy` re-render docs  that did that :unamused:

src/defaults_vs_conda-forge.rst
ocefpaf: Same as above.

src/defaults_vs_conda-forge.rst
ocefpaf: Not sure I follow. The recommendation is to install directly into an env and, in that case, `conda` should get the proper library if the channels are in order, no?

src/defaults_vs_conda-forge.rst
ocefpaf: Agreed. I did not find one easily right now nor I could force once. So I will leave this open for a future improvement.

ocefpaf:
@scopatz as our doc expert do you mind taking a look at this?

ocefpaf:
OK. Let's merge this and work from there.

PR #335, opened 2017-02-15
Update readline pin to 7.0

asmeurer:


jakirkham:
How does this compare to the `readline` pinning in `defaults`?

cc @mingwandroid @conda-forge/core

asmeurer:
I'm guessing they haven't updated it. 

We should test this, by the way. Maybe we should try building a Python package with the updated pinning first and see if it works.

jakirkham:
Done in PR ( https://github.com/conda-forge/conda-forge.github.io/pull/494 ). Thanks @asmeurer.

PR #348, opened 2017-03-03
Added Warehouse and sha256 info

artPlusPlus:
Added link to Warehouse and how to get a package's sha256.


ocefpaf:
This is good news! We should improve conda-skeleton to read that info.

PR #351, opened 2017-03-14
MNT: Close the string literal

ericdill:


jakirkham:
Ouch! 😱 Thanks @ericdill.

ericdill:
@jakirkham You got it :) 

Can you point me to the place where this script actually gets used?

jakirkham:
It gets run as part of Heroku web service and is built using this [code]( https://github.com/conda-forge/conda-forge-maintenance/ ).

Right now the script is kind of broken. I was trying to fix it in PR ( https://github.com/conda-forge/conda-forge.github.io/pull/310 ), but haven't had time to do so. If you have time to take a look, I'd definitely appreciate the extra set of eyes.

ericdill:
Gave it my :eyes: but I am awful with regexes. Seems like its worth adding a test for this before too much more tweaks happen to this script. That way it would be clear what the supported input and output of this script are. Any suggestions as to how I might contribute a test here?

croth1:
I think the plan was to break this code out of conda-forge.github.io at some point, because we might want to use it in smithy as well. I think it's time to decide on that so that we can do things properly. I agree that writing proper testing is a very good idea.

croth1:
I outsourced two crucial function for sending the pinning PRs and started writing tests here: https://github.com/croth1/conda-forge-pinning/pull/3

The general idea was not only to fix the remaining bugs, but also make it possible to use pinning information in the linter.

PR #353, opened 2017-03-15
Update recipe.rst.txt

embray:
Fixes some confusing wording regarding the SHA hash of a package.


ocefpaf:
This fixes typos so there is no need to wait. Thanks @embray!

PR #359, opened 2017-03-24
Provide info about committing re-renderings

jakirkham:
Simply explain the option to commit re-renderings with a default message.

### DIFF
--- a/src/conda_smithy.rst
+++ b/src/conda_smithy.rst
@@ -23,10 +23,18 @@ now, from the root directory of the feedstock, type:
 
 .. code-block:: shell
 
-    conda smithy rerender
+    conda smithy rerender [--commit {edit|auto}]
 
 
 and commit the changes!
+
+Optionally one can commit the changes automatically.
+To do this use the ``--commit``/``-c`` option. By default, conda-smithy
+will have git open an editor to write the commit message.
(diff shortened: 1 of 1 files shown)


jakirkham:
Could you please take a look at this @minrk?

src/conda_smithy.rst
```diff
@@ -41,6 +41,14 @@ conda smithy rerender
 
 and commit the changes!
 
+Optionally one can commit the changes automatically.
+To do this just use the `--commit`/`-c` option. By default,
+this will open an editor to make a commit. It will provide
```
minrk: Not sure if it matters, but this could be interpreted as smithy opening the editor, rather than git doing its own default behavior. One option for that is something like:

> By default, git will open an editor to write the commit message.

minrk:
👍  Minor prose nit: "just" can usually just be removed, especially in instructions.

jakirkham:
Alright filled address your comments and fixed a couple of others. Please let me know what you think.

jakirkham:
Thoughts on this @scopatz?

scopatz:
LGTM Thanks!

jakirkham:
Awesome. Thanks for taking a look guys.

PR #360, opened 2017-03-24
Pin to qt 5.6.* and pyqt 5.6.*

ocefpaf:
Not ready for merge yet. We don't have `qt 5` (nor `qt 4`) on `OS X` yet. However, it seems that `defaults` does not user `jpeg` or `libpng` in the `OS X` version of `qt 5` making it OK with `conda-forge` software stack.

I will perform some tests, confirm that, and merge this later.


jakirkham:
Why do we need to pin Qt?

ocefpaf:
`matplotlib`, `qtconsole`, adn `spyder` are OK. Thanks to @ccordoba12, @gillins, and @mingwandroid!
`conda-forge` is finally on board with `qt 5` and closer to `defaults`!

> Why do we need to pin Qt?

Future proofing and to ensure the software stack consistency. Pretty much the same reason we pin anything. Note that we still do not have `OS X` packages for `Qt 5` and, to be really safe, we should get them quickly otherwise any change on `defaults`  that adds a library with different pinning will break things on `conda-forge`.

BTW @mingwandroid we should get back to our discussion towards a consistent pinning :wink: I made the move to `icu 58.*` but there is plenty more to do.

PR #375, opened 2017-04-20
Add tick_my_feedstocks.py

pmlandwehr:
Adds a script, `tick_my_feedstocks.py` that can be run locally to check all of a given user's feedstocks to see if they are current, patch their version numbers and checksums, regenerate feedstocks, and submit pulls.

The script should work with `conda-execute`, but running it currently results in a crash. This *appears* to be an issue with `conda-execute` itself, but this needs some verification.

### DIFF
--- a/scripts/tick_my_feedstocks.py
+++ b/scripts/tick_my_feedstocks.py
@@ -511,17 +511,18 @@ def tick_feedstocks(gh_password=None,
     up_to_date_count = 0
     for feedstock in tqdm(feedstocks, desc='Checking feedstock statuses...'):
         status = feedstock_status(feedstock)
-        if status.success and status.needs_update:
-            can_be_updated.append(fs_status(feedstock, status))
-        elif not status.success:
-            status_error_dict[status.data].append(feedstock.name)
+        if status.success:
+            if status.data is None:
+                up_to_date_count += 1
+            else:
+                can_be_updated.append(fs_status(feedstock, status.data))
         else:
-            up_to_date_count += 1
+            status_error_dict[status.data].append(feedstock.name)
 
     package_names = set([x.fs.name[:-10] for x in can_be_updated])
 
     indep_updates = [x for x in can_be_updated
-                     if len(x.status.data.reqs & package_names) < 1]
+                     if len(x.sd.meta_yaml.reqs & package_names) < 1]
 
     successful_forks = list()
     successful_updates = list()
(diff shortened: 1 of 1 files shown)


pmlandwehr:
I will add that one important element of the script that hasn't yet been debugged is the creation of pulls via `pygithub` - in earlier versions I used requests for this same function, and components may have changed.

pmlandwehr:
Submitting pull requests is now officially working here, as I've gone back to `requests` from `pygithub`. Regeneration still needs further testing.

pmlandwehr:
Based on https://github.com/PyGithub/PyGithub/blob/master/github/Repository.py#L993, it seems like we can switch back from requests provided that we provide an empty `body` string in addition to the other arguments.

pmlandwehr:
Yep, that fixed it. Pulls working all over the place now.

pmlandwehr:
(Roughed out code to hit PyPI's legacy JSON API for SHAs instead of the HTML on PyPI.org, but needs testing.)

goanpeca:
Great work!

I know it is a self contained script, but it would be nice to have some tests :-p

pmlandwehr:
Sure - but where should the tests live in this case?

jakirkham:
TBH I don't think it is really that easy to test a `conda-execute` script. Especially if it is dependent on foreign APIs. FWIW the other `conda-execute` scripts here also don't have tests. I'm not against figuring out how to add tests, but I don't see that as a reasonable issue to address near term. Nor do I see it as a blocker to getting this work integrated.

Maybe the right approach would be to raise an issue at `conda-execute` to discuss coming up with a standard approach to testing scripts. That said, I'm happy to see these two efforts continue in parallel. We may find in the long run that functionality from these scripts migrates into `conda-smithy`, which is much easier target to test. In fact this very process is occurring with the pinning script.

pmlandwehr:
Just did a run, and was surprised by an apparent regression. None of the patches were actually applied this time! I'll be trying to debug it.

pmlandwehr:
Regression fixed, and have refactored the code by adding a class to manage the `meta.yaml`. This should be more useful in the long-term for better handling file and variable manipulations. Jinja variables are something of a hassle, so if folks have expertise about reverting compiled jinja templates back to their original content, that might be welcome. 

As is, the best need would be for someone to throw a local run of the script at their own feedstocks to see how well it works. Further, it might be good to specify some terminal desires before we do any kind of merge. (Perhaps instead of pushing to main, push to some feature branch so that others can get onto the TODOs?)

jakirkham:
So I'm kind of ok with this as is and would like to merge it to see what others try to do to push this forward. Especially since there is nice todo list included in the script. Over time we can try to add things to `conda-smithy` to simplify this script (which will make it easier to test @goanpeca 😉).

jakirkham:
Thoughts from @conda-forge/core?

pmlandwehr:
I'm certainly fine with a merge if we want to go. I've also pushed a few new changes to https://github.com/pmlandwehr/tick-my-feedstocks/tree/develop (note the branch). Everything still fits in one script that can be carried out with `conda-execute`, but can also potentially be bundled up with `pip`.  This should allow for the working in of some tests at some point.

pmlandwehr:
As a rider, let me not [these issues](https://github.com/pmlandwehr/tick-my-feedstocks/milestone/1) that I would consider to be the big ones before making a `0.1` release, so-to-speak. The script works fine right now, but these points should be revisited. In particular, the odd business with regeneration tokens point to temporary CI build paths rather than the official CI build paths needs to be resolved.

jakirkham:
Many of the scripts in this directory have a theme. Namely perform some operation on a feedstock locally and PR a change.

Have been thinking a bit about standardizing this procedure a bit more in `conda-smithy` so as to open the door for plugins or similar. We aren't there yet, but I think it would make these kinds of scripts simpler, more testable, and usable (both manually and automatically). Also it should make easier for people to customize and codify their workflow.

Am happy to discuss this more in some new issue over at the `conda-smithy` repo if there is interest. In the interim am more than happy to see this as a `conda-execute` script as those have been proven good for integration.

jakirkham:
Thanks again @pmlandwehr. Looking forward to playing with this. Feel free to submit tweaks as needed.

CJ-Wright:
Is there any interest in combining this with https://github.com/axiom-data-science/feedstockrot, which seems to pull version numbers from both pypi and github

pmlandwehr:
If you want to work on that, feel free. My time has gotten significantly more constrained of late. If you have any patches that you've swiped from feedstockrot that you think would be useful, I'd be glad to see them applied. All that said, the one fundamental difference I'd point out is that feedstockrot seemed to be designed around running on a small number of feedstocks. tick-my-feedstocks is intended to be run on as many feedstocks as you administrate.

CJ-Wright:
How does this compare against the stand alone package? https://github.com/pmlandwehr/tick-my-feedstocks

PR #376, opened 2017-04-21
Fix typos, add example, broaden maintainer job description

pmlandwehr:


src/recipe.rst
```diff
@@ -66,17 +66,20 @@ conda-forge.
 
 Optional: ``bld.bat`` and/or ``build.sh``
 ------------------------------------------
-In many cases, ``bld.bat`` and/or ``build.sh`` files are not required. Pure Python packages almost never need it.
+In many cases, ``bld.bat`` and/or ``build.sh`` files are not required. Pure Python packages almost never need them.
 If the build can be executed with one line, you may put this line in the ``script`` entry of the ``build`` section of
-the ``meta.yaml`` file.
+the ``meta.yaml`` file. (For instance, if a file uses `setuptools`, try ``script: python setup.py install --single-version-externally-managed --record=record.txt``.)
```
ericdill: Do you want to add a note that if you add this line then you should include ``setuptools`` in the build requirements?

ChrisBarker-NOAA:
or revisit using pip install instead? pip will omnkey-patch in setuptools if it needs to.

alternatively, use:

`python setup.py install --single-version-externally-managed --record=record.txt`

If the package doesn't use setuptools, then:

`pip install ./`

pmlandwehr:
I don't want to comment on that - my impression was that using `pip` was still generally frowned on as being not normal for conda-forge.

src/recipe.rst
ocefpaf: I believe that is "kind of" implicit there in the phrase "if a file uses `setuptools`." However, I would change file to package.
Also, being explicit is not bad too.

@pmlandwehr if you have the time for those changes let me know. I can merge this and do that later.

ocefpaf:
> I don't want to comment on that - my impression was that using `pip` was still generally frowned on as being not normal for conda-forge.

We have to move to `pip` install at some point. It is already bringing us grief with modern packages and `entrypoints`. (See latest PRs to `nbgrader` and `geopy`.) But we need another PR to implement that in the docs and the example. This one is good as-is.

ChrisBarker-NOAA:
I don't want to comment on that - my impression was that using pip was
still generally frowned on as being not normal for conda-forge.

I don't know about that.

I tried to switch all my recipes to pip -- it is simpler and works with
both setup tools and plain distutils packages -- and, in theory, with other
build systems in the future.

But I had a problem with pip being too "smart" about caching, so it
wouldn't re-build a package when I wanted it to. This was really only an
issue when developing/testing the package building, and wouldn't be an
issue when Vonda forge builds on a clean image.

So I think pip may be the way to go.

-CHB

src/recipe.rst
ocefpaf: We should update this to be `script: python -m pip install --no-deps --ignore-installed .` but let's merge this as is and I'll fix that later.

PR #392, opened 2017-05-10
defatuls pin info

ocefpaf:
@jjhelmus, @msarahan and @mingwandroid as promised today in the meeting here is an attempt to document what is used in `defaults` to help us improve compatibility between `conda-forge` and `defaults`.

The version commented right after the comma is the one available for **Linux** in `defaults`. I assumed that is the same available for all platforms (which is not always the case) and the one used for package pinning (which is also not true because `jpeg 9b` is available on `defaults` but the pinning is probably still using `jpeg 8d` there).

The only known incompatibilities that causes headaches are `jpeg` 8 vs 9 and `icu` 54.1 vc 58.*. `libpng` and `libtiff` seems OK for a while now.

PS: note that this PR does not make any functional change to the pinning script. This extra info is only to guide us towards better compatibility. When updating a pin please double check with `defaults`!

### DIFF
--- a/scripts/pin_the_slow_way.py
+++ b/scripts/pin_the_slow_way.py
@@ -25,49 +25,52 @@ import conda_smithy.github
 import conda_smithy.feedstocks as feedstocks
 
 
+# The version commented right after the comma is the one available for **Linux** in **defaults**.
+# That version is assumed to be the same in all platforms and the one used to build all packages.
+# The only known incompatibilities at the moment are jpeg 8 vs 9, and icu 54.1 vs 58.*.
 pinned = {
-          'boost': 'boost 1.64.*',
-          'boost-cpp': 'boost-cpp 1.64.*',
-          'bzip2': 'bzip2 1.0.*',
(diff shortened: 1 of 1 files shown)


jjhelmus:
@ocefpaf, thanks for putting this together.  

Updating `defaults` to use a newer `icu` is in the works but I expect it to take some time.  

As for the `jpeg` pinning, most packages in `defaults` should be using 9 in the same manner as `conda-forge`.  There are still a few packages in `defaults` that need to be re-built with the new `jpeg` version. [hdf4](https://github.com/ContinuumIO/anaconda-issues/issues/1360) is the major one that I know of that is causing issues.  I have been working on getting this fixed. If you run into any others give me a heads up and I'll try to get them fixed.

ocefpaf:
Awesome. Thanks!

The main issue with this approach is keeping it updated. My goal is to at least have some poor man's documentation on what is used in `defaults`, that way we have some guidance when updating the `conda-forge` pinning. Merging it...

PR #393, opened 2017-05-19
Latest changes from  @pmlandwehr/tick-my-feedstocks:develop

pmlandwehr:
Seems worth updating this to match [`baf91ba`](https://github.com/pmlandwehr/tick-my-feedstocks/commit/baf91ba3b549242139347c8d905a7763595e6d72) on my local `develop` branch. Notable changes include:
* more arguments for filtering the different feedstocks you're working with (primarily useful for testing)
* suppress printing warnings from beautifulsoup about not specifying an HTML parser. (I've laid off on the `lxml` requirement because this seemed to have some complications; `lxml` would be installed via conda execute but then fail to be found by the program.
* fix issue with renamed feedstock directories caused by using temporary directories instead of named directories.
* better doc strings, updated TODOS
* catching some additional exceptions, though they could be handled better.

Still pinning dependency to `py35`, though I think `py>=34` is all that's required.

### DIFF
--- a/scripts/tick_my_feedstocks.py
+++ b/scripts/tick_my_feedstocks.py
@@ -8,7 +8,6 @@
 #  - beautifulsoup4
 #  - gitpython
 #  - jinja2
-#  - lxml
 #  - pygithub >=1.29
 #  - pyyaml
 #  - requests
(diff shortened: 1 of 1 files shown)


pmlandwehr:
Just pushed a significant number of changes because of some untested issues in the last patch. Notably:
* fixed some missing commas in argparse commands causing failures
* fixed bad handlings of regexes for patching build numbers
* revised all error reporting around forking and failing to fork.

Seems to be running as expected right now, & I believe this is good for a merge.

jakirkham:
Thanks @pmlandwehr.

PR #405, opened 2017-07-20
pin flann

dougalsutherland:
flann went from 1.8.4 to 1.9.1. I don't know whether a 1.9.2 will
necessarily keep binary compatibility, but...probably? Development seems
pretty much stalled anyway.

### DIFF
--- a/scripts/pin_the_slow_way.py
+++ b/scripts/pin_the_slow_way.py
@@ -42,6 +42,7 @@ pinned = {
           'bzip2': 'bzip2 1.0.*',  # 1.0.6
           'cairo': 'cairo 1.14.*',  # 1.12.18
           'ffmpeg': 'ffmpeg >=3.2.3,<3.2.6',  # NA
+          'flann': 'flann 1.9.1',  # NA
           'fontconfig': 'fontconfig 2.12.*',  # 2.12.1
           'freetype': 'freetype 2.7',  # 2.5.5
           'geos': 'geos 3.5.1',  # 3.5.0


jakirkham:
cc @jakevdp

jakevdp:
I don't know...

jakirkham:
Could you please elaborate @jakevdp?

dougalsutherland:
For the record, based on [a search](https://github.com/search?q=org%3Aconda-forge+flann&type=Code), it seems like the recipes using flann are:

- [opencv](https://github.com/conda-forge/opencv-feedstock/blob/master/recipe/meta.yaml)
- [megaman](https://github.com/conda-forge/megaman-feedstock/blob/master/recipe/meta.yaml)
- [cyflann](https://github.com/conda-forge/cyflann-feedstock/blob/master/recipe/meta.yaml)

(plus pyflann, which matches the flann version)

jakevdp:
> Could you please elaborate @jakevdp?

I don't know what the version numbers imply, or which version numbers are or are not compatible with others.

dougalsutherland:
Looking into it a bit:

- 1.9.1 is nearly the same as 1.8.5: looking at the source, it's just a version number bump and the addition of one dummy `delete` operator. So they're ABI-compatible except for the change in library name.
- https://github.com/lvc/abi-compliance-checker says that 1.8.5 is incompatible with 1.8.4 (51 binary compatibility problems, 224 warnings). Note that 1.8.5 doesn't name its libraries properly; it's `libflann.so.1.8.4`.
- 1.8.1 was only 50% binary compatible with 1.8.4 (!).

So it seems like it should be pinned to the minor release.

jakirkham:
Given @dougalsutherland recent comment, does this clarify anything for you, @jakevdp?

jakevdp:
Thanks for the ping. What effect would pinning the version here have? This is the github-pages repo, yes?

jakirkham:
So this script has been used to update pinnings in feedstocks in batch. Why it lives here is a bit of a historical oddity at this point. The short answer is we used to use CIs on this repo to do some work like this in the past.

For now, this is more where we document these pinnings (due to some bugs in the service that handles these pinning updates), but we plan to change that as part of the `conda-build` 3 transition in some manner YTBD.

jakevdp:
OK – it seems like this is fine then.

jakirkham:
Thanks all.

PR #407, opened 2017-07-20
pin curl

dougalsutherland:
Curl apparently hasn't added a new symbol [in years](https://abi-laboratory.pro/tracker/timeline/curl/), but maybe they will next time!

### DIFF
--- a/scripts/pin_the_slow_way.py
+++ b/scripts/pin_the_slow_way.py
@@ -41,6 +41,7 @@ pinned = {
           'boost-cpp': 'boost-cpp 1.64.*',  # NA
           'bzip2': 'bzip2 1.0.*',  # 1.0.6
           'cairo': 'cairo 1.14.*',  # 1.12.18
+          'curl': 'curl >=7.44.0,<8'  # 7.54.1
           'ffmpeg': 'ffmpeg >=3.2.3,<3.2.6',  # NA
           'fontconfig': 'fontconfig 2.12.*',  # 2.12.1
           'freetype': 'freetype 2.7',  # 2.5.5


jakirkham:
Yeah, I'm leery of pinning it too tightly on the upper bound. As noted, it really hasn't added/changed symbols. Plus fixes in `curl` tend to be CVEs. So I wouldn't want it to be too hard to get those fixes in the stack.

dougalsutherland:
Yeah, that definitely makes sense. I just figured I'd put it here for completeness.

jakirkham:
Maybe we can make the upper bound 8 then. Thoughts?

jakirkham:
LGTM. @conda-forge/core?

scopatz:
Thanks for putting this in @dougalsutherland. Adding a symbol isn't a problem, only removing a symbol would be.  So why would we need a pin here?

jakirkham:
TBH the first version of `curl` we added to `conda-forge` was 7.48. So everything has been built with the expectation of having this new symbol.

dougalsutherland:
The one thing it does is protect against people with very old versions of curl installed from `defaults`: `7.{26,30,38,42.1,43}` are still available there.

But I don't think it's a big deal, and maybe it's not worth the hassle. I only sent the PR because I was sending pinning PRs for all the C libraries I used then anyway.

scopatz:
In that case, the minimal pin makes sense, but the maximal pin seems like it is too much right?

dougalsutherland:
Well, a curl 8 would most likely break stuff, and this would protect against that. Since curl is apparently willing to add symbols in minor revisions, a minor revision might _also_ break stuff, but we can't really protect against that without updating the upper bound on the pin at every release.

scopatz:
Why do you think a curl 8 will most likely break backwards compatibility?  Please excuse my ignorance.

dougalsutherland:
I just found this: https://curl.haxx.se/libcurl/abi.html. Apparently curl began with version 7.0, so a version 8 seems pretty unlikely. :)

jakirkham:
Based on their discussion about [version numbers]( https://curl.haxx.se/docs/versions.html ), it suggests that when the major number changes they will break things. Also we have generally assumed software uses semantic versioning unless they explicitly say otherwise.

scopatz:
Alright, the upper limit is fine then.

jakirkham:
Thanks @scopatz. Comments from others?

mwcraig:
I'm increasingly coming around to the view that this kind of "defensive" pinning is a good idea.

jakirkham:
Given that, does this pinning seem ok or do you have different thoughts, @mwcraig?

mwcraig:
Oops, thought I was being clear before. I'm  👍  on keeping the <8 pinning.

jakirkham:
Thanks everyone!

PR #409, opened 2017-07-20
pin json-c

dougalsutherland:
0.12 and 0.12.1 are the same, per [the tracker](https://abi-laboratory.pro/tracker/timeline/json-c/), so assume that 0.12.2 would be too.

cc @conda-forge/json-c


jakirkham:
They definitely are breaking stuff in `master`, but agree that it will probably not be in 0.12.x then.

PR #410, opened 2017-07-20
pin lzo

dougalsutherland:
I don't know about binary compatibility on this one, since it's not on the tracker. There's actually a 2.10 release, though the PR for it (https://github.com/conda-forge/lzo-feedstock/pull/2) hit some build error that nobody ever debugged apparently. 2.09 is presumably fine for now.

cc @conda-forge/lzo

### DIFF
--- a/scripts/pin_the_slow_way.py
+++ b/scripts/pin_the_slow_way.py
@@ -62,6 +62,7 @@ pinned = {
           'libsvm': 'libsvm 3.21|3.21.*',  # NA
           'libtiff': 'libtiff >=4.0.3,<4.0.8',  # 4.0.6
           'libxml2': 'libxml2 2.9.*',  # 2.9.4
+          'lzo': 'lzo 2.*',  # 2.06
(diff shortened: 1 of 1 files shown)


jakirkham:
Would raise it [here]( https://github.com/lvc/upstream-tracker/issues ) as well.

dougalsutherland:
lzo actually looks stable [since 2005](https://abi-laboratory.pro/tracker/timeline/lzo/index.html), so it should be fine without being pinned.

jakirkham:
Maybe `2.*` then. Just in case `3.*` means API/ABI breaks.

dougalsutherland:
Sure, good idea; updated the PR to `2.*`.

jakirkham:
Thoughts @rmax?

rmax:
I second pinning to `2.*`.

jakirkham:
Thanks @dougalsutherland and @rmax.

PR #411, opened 2017-07-20
pin protobuf

dougalsutherland:
The [tracker](https://abi-laboratory.pro/tracker/timeline/protobuf/) says 3.3.* is compatible, but incompatible with 3.2.*.

If https://github.com/conda-forge/staged-recipes/pull/3087 merges, might have to pin both? (Or maybe `protobuf` will depend on `libprotobuf`? It does in defaults.)

cc @conda-forge/protobuf

### DIFF
--- a/scripts/pin_the_slow_way.py
+++ b/scripts/pin_the_slow_way.py
@@ -75,6 +75,7 @@ pinned = {
           'pango': 'pango 1.40.*',  # 1.40.3
           'pixman': 'pixman 0.34.*',  # 0.34.0
           'proj4': 'proj4 4.9.3',  # 4.9.2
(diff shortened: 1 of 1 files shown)


dougalsutherland:
Protobuf has since released an incompatible version 3.4, which is causing at least some problems: https://github.com/conda-forge/caffe-feedstock/issues/31. Updating this PR to that.

dougalsutherland:
Gentle ping. :)

dopplershift:
I'm ok with it from a protobuf standpoint, but know nothing about how pinning works to evaluate this PR.

jakirkham:
From a brief search of the org, some people have already started using this version. The ones that haven't either failed to pin `protobuf` or are on some ancient version. Let's go ahead with this.

PR #412, opened 2017-07-20
pin snappy

dougalsutherland:
Pins to current most recent version in conda-forge, 1.1.4.

1.1.6 has a pending PR: https://github.com/conda-forge/snappy-feedstock/pull/11. It apparently has a slight ABI incompatibility, per [the tracker](https://abi-laboratory.pro/tracker/timeline/snappy/). (1.1.5 was broken in between.)

cc @conda-forge/snappy


dougalsutherland:
Since https://github.com/conda-forge/snappy-feedstock/pull/11 merged, switched to pin to 1.1.6. I suppose we might want to keep it at 1.1.4 anyway for compatibility with `defaults`, though.

dougalsutherland:
(Updated comment: defaults has 1.1.6 now.)

jakirkham:
Thanks @dougalsutherland.

PR #418, opened 2017-07-25
update openblas pin

ocefpaf:
I don't see an `openblas 0.2.20` in `defaults` yet. I am assuming the "pinning" there is still `openblas 0.2.19`.


ocefpaf:
@isuruf let's keep this one open until we fix the `scipy` builds with the latest `blas`.

jschueller:
what's the status on this ?

jschueller:
ping @ocefpaf

jschueller:
hey @sofianehaddad it seems I cannot apply the patch to openblas 0.2.19 as the offending commit was not part of that release, maybe it's another issue ?

scripts/pin_the_slow_way.py
```diff
@@ -76,8 +76,8 @@
           'ncurses': 'ncurses 5.9',  # 5.9
           'netcdf-cxx4': 'netcdf-cxx4 4.3.*',  # NA
           'netcdf-fortran': 'netcdf-fortran 4.4.*',  #
+          'openblas': 'openblas 0.2.20|0.2.20.*',  # 0.2.19
```
ocefpaf: @isuruf is this OK to merge?

isuruf: Yes, it is. Maybe we want to add a comment that the runtime requirement can be relaxed.

ocefpaf: I am on the fence about that. We don't support that "officially" b/c (a) the pinning bot is dead, (b) even when active it could not do that.

It is an advanced technique that a comment might encourage it, and I'd rather leave that to those who a certain about what they are doing.

isuruf: Okay. I thought it would be better, since we do the same thing for numpy.

ocefpaf:
Once we have a 'pin differently policy' yes, it will be better.

The numpy case is an exception b/c it affects our CI matrix.

jschueller:
Do you plan on proposing prs to update the pinning in feedstocks ?

PR #421, opened 2017-08-03
update numpy versions

ocefpaf:
Some changes to reflect the `numpy 1.8` packages. Only `matplolib` and `pandas` are taking advantage of `numpy 1.7` and, b/c it would need to be patched to build on Python 3.5 and 3.6, I decided to just drop it from the docs. That does not mean people cannot use it if they want and they know what they are doing.

@jjhelmus or @isuruf can you take a quick look at this?

Ping @pkg.

### DIFF
--- a/src/meta.rst
+++ b/src/meta.rst
@@ -73,22 +73,24 @@ A full description of selectors is
 
 Building Against NumPy
 ----------------------
-If you have a package which links\* against numpy you need to build against the oldest possible version of numpy that is forwards compatible.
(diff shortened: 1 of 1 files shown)


kynan:
Is this way still able to prevent incompatible NumPy ABI versions?

ocefpaf:
Yep. That is the goal ;-)

kynan:
OK, I'll apply that to the other NumPy dependent packages I maintain then!

PR #433, opened 2017-08-29
Updated numpy matrix suggestion

ocefpaf:
I was too optimistic that the would add older `numpy`s for Windows quickly :smile: 

I am adjusting this until that is rectified.

@isuruf can you take a quick look?

### DIFF
--- a/src/meta.rst
+++ b/src/meta.rst
@@ -80,18 +80,26 @@ At the moment these are the oldest available ``numpy`` versions in conda-forge t
 .. code-block:: yaml
 
     build:
-      - numpy 1.8.*  # [not (win and py36)]
+      - numpy 1.8.*  # [not (win and (py35 or py36))]
+      - numpy 1.9.*  # [win and py35]
       - numpy 1.11.*  # [win and py36]
     run:
-      - numpy >=1.8  # [not (win and py36)]
+      - numpy >=1.8  # [not (win and (py35 or py36))]
+      - numpy >=1.9  # [win and py35]
       - numpy >=1.11  # [win and py36]
 
 We will add older versions for ``Python 3.6`` on Windows soon.
-Note that you still need to respect the package minimum supported version of ``numpy``!
-That means you cannot use ``numpy 1.8`` if the project requires at least ``numpy 1.9``.
+That will allow us to simplify the minimum ``numpy`` to ``1.8`` across all platforms and Python versions.
+
+CAVEAT: note that you still need to respect minimum supported version of ``numpy`` for the package!
+That means you cannot use ``numpy 1.8`` if the project requires at least ``numpy 1.9``,
+adjust the minimum version accordingly!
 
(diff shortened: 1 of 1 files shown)


ocefpaf:
@kain88-de you've been an early adopter of this too :smile: 
Do you mind taking a look at the text and see if that is clear enough?

msarahan:
1.9 is the oldest we're going to support.  I have 1.9 builds on the c3i_test channel as of this morning: https://anaconda.org/c3i_test/numpy/files

ocefpaf:
Cool. That is already a big help.

Just curious, is the reason to not add 1.8 the [gzip patch](https://github.com/conda-forge/numpy-feedstock/blob/numpy18/recipe/meta.yaml#L12)?  Or are there other complications on Windows that does not make it worth it?

msarahan:
I kept the gzip patch in 1.8, but wasn't sure if it was necessary in 1.9.  There are no patches applied in the conda-forge 1.9 recipe: https://github.com/conda-forge/numpy-feedstock/blob/numpy19/recipe/meta.yaml

ocefpaf:
> but wasn't sure if it was necessary in 1.9.

No, it is not necessary. Only for 1.8. I guess that my question was not well formed.

The reason to not support numpy 1.8 was the patch (or patching in general to make it work with py35 and py36) or other reasons?

msarahan:
Ah.  There were many patches required for 1.8.  Mostly all for VS2015.  I started to backport them, but it just wasn't worth the effort.

msarahan:
PS: I would also rather have a single version than all this selector junk.  I could have had 1.8 for everything but win py3.5+, but I think 1.9 is good enough, and having just 1.9 is so much less headache.

ocefpaf:
> Ah. There were many patches required for 1.8. Mostly all for VS2015. I started to backport them, but it just wasn't worth the effort.

Got it. I'll update the docs as soon as the new packages land on `defaults`. For some a minimum of `1.9` makes more sense every day :smile:

ocefpaf:
> PS: I would also rather have a single version than all this selector junk. I could have had 1.8 for everything but win py3.5+, but I think 1.9 is good enough, and having just 1.9 is so much less headache.

You just missed the opportunity to advertise conda-build 3 :wink:

msarahan:
Yeah, cb3 would be great for this actually.  Having the selector in just conda_build_config.yaml would make it feasible to support 1.8 and 1.9 more easily.  Meh...  Maybe I'll go back and build 1.8 too, now that you mention it.

ChrisBarker-NOAA:
Looks like 1.9 was released three years ago -- I think we're good. :-)

-CHB

kain88-de:
Looks good to me

ocefpaf:
@msarahan please let me know when the Windows pkg for numpy 1.9 on Python 3.6 and 3.5 land on `defaults` so I can update this. In short term I guess we can live with:

```yaml
build:
  numpy 1.9.*
run:
  numpy >=1.9
```

A long term solution is to using `conda-build 3` and a `conda_build_config.yaml` to control that.

PR #437, opened 2017-09-11
Adding nettle pinning

Cashalow:
Following https://github.com/conda-forge/staged-recipes/pull/3884#discussion_r137950866 I am updating this file for nettle pinning.

ABI report: https://abi-laboratory.pro/tracker/timeline/nettle/


dougalsutherland:
It looks like:
- `nettle` isn't in defaults, so the comment should be `NA` instead of `3.3.0`.
- The version number is actually `3.3`, not `3.3.0`, so the pin should be `3.3|3.3.*`.

Cashalow:
Ok thanks for the input !

jakirkham:
Added a link to the ABI report in the OP.

jakirkham:
Thanks @Cashalow and @dougalsutherland.

PR #438, opened 2017-09-11
update snappy pin

dougalsutherland:
Update the snappy pin (just added today in #412...) to `1.1.6|1.1.7`: their ABI is [totally compatible](https://abi-laboratory.pro/tracker/timeline/snappy/), and though we don't have 1.1.7 in conda-forge yet we will soon: https://github.com/conda-forge/snappy-feedstock/pull/14.


jakirkham:
Thanks @dougalsutherland.

PR #439, opened 2017-09-13
Update boost pin

jschueller:


jschueller:
is the pinning bot still down ?

jakirkham:
Yes. 😞 Tried to fix it a while back with PR ( https://github.com/conda-forge/conda-forge.github.io/pull/310 ), but have lacked the time to complete it. If you are interested, would happily look at a new PR to fix the bot.

jschueller:
@jakirkham how can I test that script locally to debug ? it seems tied to the admin account. Can you show me a feedstock that it cannot update ?

jakirkham:
Yeah I don't think you can test it locally that easily, which is definitely a problem. Though it looks like you figured it out.

jschueller:
yes I isolated the few relevant bits

jakirkham:
Went back and realized that `boost` actually broke ABI in the patch release. We need to update this to 1.65.1. All the SONAMEs changed and `libboost_regex` got some new symbols and removed some symbols as well.

ref: https://abi-laboratory.pro/tracker/objects_report/boost/1_65_0/1_65_1/report.html

jschueller:
https://github.com/conda-forge/conda-forge.github.io/pull/455

PR #454, opened 2017-09-24
Update libpng pinning

jakirkham:
Appears that libpng is stable beginning with 1.6.22, but ends up breaking ABI in version 1.6.31 from this year. So adjust the lower and upper bounds of the pinning to match this.

ref: https://abi-laboratory.pro/tracker/timeline/libpng/

cc @conda-forge/core @conda-forge/libpng


scopatz:
Seems good to me!

PR #457, opened 2017-10-02
Force latest git with feedstock updates

jakirkham:
Seems that recently `conda` has been picking an older version of `git` ( maybe to have a newer version of `zlib` :/ ). That is the only notable change between the feedstocks' script failing and working. So try forcing the newer `git`.

Passing: https://travis-ci.org/conda-forge/conda-forge.github.io/jobs/281593290#L522
Failing: https://travis-ci.org/conda-forge/conda-forge.github.io/jobs/279785043#L522

### DIFF
--- a/.ci_scripts/update_feedstocks_repo
+++ b/.ci_scripts/update_feedstocks_repo
@@ -5,7 +5,7 @@
 set -e
 
 if [ -n "$GH_TOKEN" ]; then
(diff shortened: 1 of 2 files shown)


jakirkham:
Well was going to try to fix this enough to do one last run before we remove this in favor a webservice that handles feedstocks changes. However it seems this is still broken. Besides have gone through an updated all the feedstocks since the last successful run through the webservice. So I guess we can just get rid of this since it is finally broken and now unnecessary.

PR #466, opened 2017-10-15
Drop feedstocks listing

jakirkham:
Removes the current feedstocks listing as it has been moved to the feedstocks repo. Also no longer generates the feedstocks listing as part of the CI build. Instead simply builds the docs here as per usual. As the feedstocks listing is now handled by a webservice and stored in the feedstocks repo ( https://github.com/conda-forge/conda-forge-webservices/pull/142 ), there should be no issue dropping either of these.


jakirkham:
Going to try this out to see if this will work.

PR #467, opened 2017-10-15
Use absolute links in feedstocks template

jakirkham:
To make sure that everything links properly from the feedstocks listing, even if it is not next to the relevant resources, change all of the URLs in the feedstocks template to be absolute. This way they are guaranteed to point back to right places.

### DIFF
--- a/feedstocks.html.tmpl
+++ b/feedstocks.html.tmpl
@@ -15,7 +15,7 @@
     <link href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.min.css" rel="stylesheet">
 
     <!-- Custom CSS -->
-    <link href="css/theme.css" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


feedstocks.html.tmpl
```diff
@@ -39,27 +39,27 @@
                 <button type="button" class="navbar-toggle" data-toggle="collapse" data-target=".navbar-main-collapse">
                     <i class="fa fa-bars"></i>
                 </button>
-                <a class="navbar-brand page-scroll" href="./#page-top">
+                <a class="navbar-brand page-scroll" href="https://conda-forge.github.io/#page-top">
                     <i><img src="img/anvil_white.png" alt="An anvil" style="height: 1em;"></i>  <span class="light">conda-forge</span>
```
jakirkham: Missed fixing this link to be absolute. Correcting it in PR ( https://github.com/conda-forge/conda-forge.github.io/pull/468 ).

PR #476, opened 2017-10-31
update pinning comment about what defaults uses for protobuf

dougalsutherland:
Forgot to change this when I updated #411; sorry about that. Just to keep the comments marginally correct. :)

### DIFF
--- a/scripts/pin_the_slow_way.py
+++ b/scripts/pin_the_slow_way.py
@@ -80,7 +80,7 @@ pinned = {
           'pango': 'pango 1.40.*',  # 1.40.3
           'pixman': 'pixman 0.34.*',  # 0.34.0
           'proj4': 'proj4 4.9.3',  # 4.9.2
-          'protobuf': 'protobuf 3.4.*',  # 3.2.0
+          'protobuf': 'protobuf 3.4.*',  # 3.4.0
           'pyqt': 'pyqt 5.6.*',  # 5.6.0
           'qt': 'qt 5.6.*',  # 5.6.2
           'readline': 'readline 6.2*',  # 6.2


scripts/pin_the_slow_way.py
```diff
@@ -80,7 +80,7 @@
           'pango': 'pango 1.40.*',  # 1.40.3
           'pixman': 'pixman 0.34.*',  # 0.34.0
           'proj4': 'proj4 4.9.3',  # 4.9.2
-          'protobuf': 'protobuf 3.4.*',  # 3.2.0
+          'protobuf': 'protobuf 3.4.*',  # 3.4.0
```
jakirkham: So this is what `defaults` is using now?

dougalsutherland:
It's at least the newest version available in `defaults`. I didn't check which version any packages actually link to....

jakirkham:
@mingwandroid, does this match up with `defaults` or is a different `protobuf` version in use currently?

mingwandroid:
@jjhelmus is your go-to guy for all your `protobuf` enquiries. I think he's been working on it recently.

jakirkham:
Friendly nudge @jjhelmus. 😉

jjhelmus:
`defaults` is using protobuf 3.4.x on linux-64 and 3.2.x on the other platforms at the moment.  I've been working on getting 3.4.1 package out for all platforms but a failing unit test with the GCC 7.2.0 toolchain is holding this up.  

Having conda-forge pin protobuf 3.4.* is a big :+1: from me, `defaults` will be there soon.

jakirkham:
Good to hear. Though this PR was just trying to fix the comment on what `defaults` is pinned to. My bad if we missed you on the previous PR. Should we be saying `3.2.*` or `3.3.*` for `defaults` then?

jjhelmus:
In `defaults` protobuf is only used by the `caffe`, and `tensorflow` packages which do not use a uniform pinning.  I'm working on getting packages ready for both of these which use `protobuf` 3.4.x.

PR #482, opened 2017-11-08
pin hdf5

ocefpaf:
I guess it is time to move forward :wink: 

@jjhelmus what do you think? (You know more about hdf5 than me :stuck_out_tongue:)

### DIFF
--- a/scripts/pin_the_slow_way.py
+++ b/scripts/pin_the_slow_way.py
@@ -53,7 +53,7 @@ pinned = {
           'glpk': 'glpk 4.61|4.62',  # NA
           'gmp': 'gmp 6.1.*', # 6.1.0
           'harfbuzz': 'harfbuzz 1.3.4',  # 0.9.39
-          'hdf5': 'hdf5 1.8.18|1.8.18.*',  # 1.8.17
+          'hdf5': 'hdf5 1.10.1',  # 1.8.17
           'icu': 'icu 58.*',  # 54.1
           'jpeg': 'jpeg 9*',  # 9b (but defaults is probably using 8d)
           'json-c': 'json-c 0.12|0.12.*',  # NA


jjhelmus:
hdf5 1.10.x introduced a new "sub-format" for HDF5 files, the version 3 superblock.  It adds a number of nice features but these changes do break compatibility.  It is possible to create a HDF5 file with hdf5 1.10.x which cannot be read with hdf5 1.8.x.  Because of this it may be helpful to maintain two builds of packages which require hdf5, one with 1.8.x to ensure compatibility and another with 1.10.x.  

I haven't been keeping up with how the new sub-format effect NetCDF4.  I recall some initial issues with 1.10.x , see Unidata/netcdf-c/#250 for details.  From my reading netcdf4 >=4.4.1 supports hdf5 1.10.x in a backwards compatible manner.  Both `defaults` and `conda-forge` provide 4.4.1 or later so this should be fine.

ocefpaf:
> I haven't been keeping up with how the new sub-format effect NetCDF4.  I recall some initial issues with 1.10.x , see Unidata/netcdf-c/250 for details.  From my reading netcdf4 >=4.4.1 supports hdf5 1.10.x in a backwards compatible manner.  Both `defaults` and `conda-forge` provide 4.4.1 or later so this should be fine. 

That was my concern as well but now I see that https://github.com/Unidata/netcdf-c/issues/250 is open but "solved."

BTW, people pointed to me repeatedly that wheels are using `libnetcdf 4.4.1.1` and `hdf5 1.10.1` just fine, so I guess we should update it here too.

Thanks for the detailed explanation!

jakirkham:
We should probably make a list of things to update and make sure we got them all.

tacaswell:
wait, h5py wheels are built with 1.10?!

ocefpaf:
> wait, h5py wheels are built with 1.10?!

@tacaswell maybe I misinterpret the https://github.com/h5py/h5py/issues/880 thread but it seems to me that Windows wheels are built with `1.10.1`.

jjhelmus:
> wait, h5py wheels are built with 1.10?!

The hdf5-2.7.1 linux wheel appear to be. I have not checked the other platforms.

```
$ unzip -l h5py-2.7.1-cp35-cp35m-manylinux1_x86_64.whl | grep "libhdf5"
   185880  2017-09-01 06:19   h5py/.libs/libhdf5_hl-6491d8e2.so.10.1.1
  3486080  2017-09-01 06:19   h5py/.libs/libhdf5-6675071d.so.10.2.1
```

PR #490, opened 2017-11-23
pin poppler

ocefpaf:
Defaults added `poppler 0.60.1` and we never had that pinned b/c `conda-forge` was the only source. Now we have to b/c it is breaking packages that links to `poppler`.

Same story with `libiconv`/


scripts/pin_the_slow_way.py
```diff
@@ -82,6 +82,7 @@
           'openssl': 'openssl 1.0.*',  # 1.0.2k
           'pango': 'pango 1.40.*',  # 1.40.3
           'pixman': 'pixman 0.34.*',  # 0.34.0
+          'poppler': 'poppler 0.61.*'  # 0.60.1
```
ocefpaf: The current version on conda-forge is `0.52.0` but I am updating that now to `0.61.1`.

PR #511, opened 2018-01-13
pin zeromq to 4.2.1

minrk:
An ABI-specifying bug in the zeromq recipe means we need to pin zeromq to the patch level until a fix is issued.

See related discussion in https://github.com/conda-forge/pyzmq-feedstock/pull/25 and https://github.com/conda-forge/zeromq-feedstock/issues/25


ocefpaf:
Thanks @minrk I'll start fixing the pinning in the feedstock over the weekend to avoid the Travis-CI backlog.

PR #521, opened 2018-02-05
Fix last item border.

Carreau:
Removing bottom border from all elements to avoid double border make
sens; except for the last !

--- 
One question though why `border-bottom: 1px solid rgba(255,255,255,.1);` (white, almost invisible) and not `none`?


ocefpaf:
Thanks @Carreau!

PR #523, opened 2018-02-08
Add an onerror handler to shutil.rmtree calls in tick_my_feedstocks.py

tkelman:
Should avoid errors such as
`PermissionError: [WinError 5] Access is denied: 'C:\\Users\\akelman\\AppData\\Local\\Temp\\tmpu3fwwzf4\\botocore-feedstock\\.git\\objects\\pack\\pack-75b2092341bff03adcd15b32b752762b5801440f.idx'`

### DIFF
--- a/scripts/tick_my_feedstocks.py
+++ b/scripts/tick_my_feedstocks.py
@@ -97,6 +97,7 @@ from pkg_resources import parse_version
 import re
 import requests
 import shutil
+import stat
 import tempfile
 from tqdm import tqdm
 import yaml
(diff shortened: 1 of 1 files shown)


tkelman:
I'd also kind of like to move the `conda_smithy` imports inside the `regenerate_fork` function so it's an optional dependency that you can avoid if you use `--no-regenerate`, but let me know if anyone has any preference for doing that in this PR or a separate one.

pmlandwehr:
Thanks for this - might also want to check how it squares with [the `develop` branch of the *other* `tick-my-feedstocks`](https://github.com/pmlandwehr/tick-my-feedstocks/tree/develop) I should really seem about merging those changes back into here at some point...

tkelman:
Right, saw that, opened https://github.com/pmlandwehr/tick-my-feedstocks/pull/19 with basically the same change.

For another couple-line but only partially related change, would you prefer it to be added to these existing PR's or done separately?

PR #531, opened 2018-02-15
optimize feedstocks search

### LINKED ISSUE #529: feedstocks page hangs when using filter
I noticed that the filter items search bar at https://conda-forge.org/feedstocks/ causes my browser to hang on the first letter I type into the search bar. I'd love to help improve this functionality for the site if deemed necessary.

djm158:
fixes #529

### DIFF
--- a/css/theme.css
+++ b/css/theme.css
@@ -82,16 +82,6 @@ a:focus {
     margin-bottom: 0;
     font-family: Montserrat,"Helvetica Neue",Helvetica,Arial,sans-serif;
     color: black;
-    transition: all 0.1s ease;
-}
-
-
-.list-group-item.search-hidden {
-  height:0;
-  padding-top:0;
-  padding-bottom:0;
-  opacity:0;
-  border:none;
 }
 
 .list-group-item:not(:last-child){
--- a/feedstocks.html.tmpl
+++ b/feedstocks.html.tmpl
@@ -131,6 +131,26 @@
 
     </script>
     <script>
+        // Returns a function, that, as long as it continues to be invoked, will not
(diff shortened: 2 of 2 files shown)


jakirkham:
@Carreau, would you have time to give this a quick look?

Carreau:
Looks good, though changing the inner-html may make css transition to not work, and in case we do not want transitions, we can just remove this extra css.

We also can (and should ?) debounce the search to have less querries/seconds.

djm158:
@Carreau are you saying remove the css transition in the class `.list-group-item`? Also should I wrap the callback function for the search field in a `__.debounce`? 

e.g.:

```js
$('#searchfield').show().on('change keyup keypress paste', __.debounce(function() {
    // do search
}, 200))
```

Carreau:
yes, you could just remove the 
```css
//theme.css:86
.list-group-item.search-hidden {
   height:0;
   padding-top:0;
   padding-bottom:0;
   opacity:0;
   border:none;
 }
```

And `transition: all 0.1s ease;`.

`_` is not available on the page, so you might need to create your own debounce function.

djm158:
I see, should be easy enough. We can probably just do something similar to the way underscore does their [debounce](http://underscorejs.org/#debounce) function.

jakirkham:
Thoughts @Carreau?

Carreau:
I say let's try !

jakirkham:
Thanks all. :)

jakirkham:
Just wanted to say, this was a really nice fix. It's very responsive now. Thanks both of you for helping improve it.

Issue #530, opened 2018-02-14
improve documentation for building this repo


djm158:
It would be helpful if the README contained info for building the documentation in this repo locally. Something like:

## Contributing 

```sh
git clone https://github.com/conda-forge/conda-forge.github.io.git
cd conda-forge.github.io
```

### Development environment

install [conda](https://conda.io/docs/user-guide/install/index.html#)

```sh
conda install -n root -c conda-forge conda-smithy
```

### Build the documentation

install [shpinx](http://www.sphinx-doc.org/en/master/)

```sh
conda install sphinx
```

Then navigate to the `src` directory.

Windows:

```sh
./make.bat
```

UNIX:

```sh
make
```

### Build the html for http://conda-forge.org

```sh
python scripts/generate_html.py --html-source "."
```

then navigate to `index.html` in the root of your `conda-forge.github.io`

ChrisBarker-NOAA:
Shouldn’t that be:

conda install sphinx 

:-)

djm158:
haha updated 😊

jakirkham:
Sure a PR would be welcome.

I think it is actually simpler than the steps above as we no longer generate the feedstock listing in this repo. So it's just install `sphinx` and run `make` (Windows normally checks some executable extensions for you).

djm158:
will `make` generate the html for `feedstocks.html.tmpl`? If so could you point me where to look? It appears that `make` generates the documentation pages.

jakirkham:
Sorry are you wanting to generate the feedstock listing or the docs?

The feedstock listing code has largely been moved out of this repo (with the exception of the template) into this [webservices' function]( https://github.com/conda-forge/conda-forge-webservices/blob/1cbc6580ddf155f86c63ee8cbcae80b84900aee7/conda_forge_webservices/feedstocks_service.py#L16 ).

Correct, `make` should just generate the docs.

djm158:
Ohhh okay that makes sense. I was just trying to generate the listing to work on #531.

jakirkham:
Yeah, sorry, it's not very dev friendly ATM as it requires 3 repos to build currently (webservices, feedstocks, and the webpage repo). Basically we list the contents of the feedstocks repo and use that to fill out the webpage template. It lives at webservices as we do live updating on Heroku when new feedstocks are made.

If you have any ideas about how we can make it more dev friendly, definitely would be open to suggestions.

jakirkham:
[...]

PR #540, opened 2018-03-01
better docs about src/, docs/ folders (#530)

dougalsutherland:


jakirkham:
Wonder if we should add a GitHub contributing template for PRs to make this more visible.

src/README.md
```diff
@@ -0,0 +1,3 @@
+To build the docs here locally: make sure you have Sphinx installed (`conda install sphinx`), then run `make html` (Unix) or `./make.bat html` (Windows). This will create the output files in `_build/html`.
```
jakirkham: Probably should add `cd src`.

PR #541, opened 2018-03-01
update_docs: remove outdated files from docs/

dougalsutherland:
The `docs/` dir contains some outdated files: old versions of jquery in `docs/_static`, and `docs/_sources` has old `guidelines.txt`/`index.txt` files (instead of the current ones like `guidelines.rst.txt`). This is because `update_docs` currently doesn't delete old files from `docs`, just builds the new ones on top of it.

This PR changes the `update_docs` script so that it removes the old content first. The one hacky thing it does is to keep the `README` file around; there's probably a better way to do that....


ocefpaf:
Thanks!

Issue #565, opened 2018-04-14
"conda skeleton pypi" Meta-Mega-Issue


leycec:
Per @CJ-Wright's [sagacious advice](https://github.com/conda-forge/staged-recipes/pull/5574#issuecomment-381300150) at [my first conda-forge recipe](https://github.com/conda-forge/staged-recipes/pull/5574), this large-scale issue collects various interrelated ~~critiques~~ concerns of current documentation on the `conda skeleton pypi` command.

## Isn't All of This conda-build's Fault?

*Maybe.* Because of that likelihood, @CJ-Wright wisely suggested I open this issue at [the core `conda-build` tracker](https://github.com/conda/conda-build/issues) instead.

I unwisely ignored this suggestion. It's not necessarily clear to me that anything is explicitly broken *or* in need of immediate repair in the `conda skeleton pypi` command. Moreover, unconditionally rewriting the output produced by that command in the manner detailed below could be a Bad Thing™ in edge cases that I (as a sleep-deprived cottage neckbeard) am unaware of. Certainly, rewriting the `requirements/host:` list into a `requirements/build:` list would *definitely* be a Bad Thing™ in the general case of conda-build 3 support.

Many of the concerns listed below are probably specific to conda-forge. So, I don't know. I'm dumping them here – but sympathize with @CJ-Wright's preferred solution of improving the `conda skeleton pypi` command itself. Maybe we can do both? That is, maybe we can both improve conda-forge documentation (*which seems like low-hanging fruit*) and also improve `conda skeleton pypi` (*which really doesn't*).

Here we go, folks. Thanks again to @carlodri, @CJ-Wright, and @jakirkham for their tireless efforts.

## How Bad Could "conda skeleton pypi" Documentation Be, Anyway?

**Pretty bad,** frankly. The [conda-forge documentation](https://conda-forge.org/docs/recipe.html#getting-started) on this command is scant at best:

> If it is a python package you can generate a skeleton as a starting point with `conda skeleton pypi your_package_name`. You do not have to use skeleton, and **the recipes produced by skeleton will need to be edited.**

[...]

PR #569, opened 2018-04-14
tweak getting-started docs

dougalsutherland:
De-emphasize usage of `conda skeleton pypi` in favor of modifying our `example` recipe, and list a few specific things to be changed if you do use it.

Helps with #565.


jakirkham:
Thanks @dougalsutherland. :)

PR #582, opened 2018-05-11
Updated hash docs to reflect new PyPi website.

moorepants:
pypi.org is now switched to the new warehouse software and only provides a SHA256 hash on the download page. Updating docs to reflect.


jakirkham:
Thanks @moorepants. :)

PR #583, opened 2018-05-11
Fixes link to conda-forge team members

jrbourbeau:
When reading through the "Creating Conda Recipes" documentation, I ran across a broken link to the conda-forge team members. This PR simply fixes the link.


jrbourbeau:
@jakirkham you're everywhere! 

> it probably only works if you are already a conda-forge member

Oh, I see. I hadn't considered that possibility. 

At any rate, the updated link should work for both conda-forge members and non-members (such as myself).

PR #598, opened 2018-06-07
Governance document

scopatz:
Probably long overdue, but I hope it will help!


src/governance.rst
```diff
@@ -0,0 +1,172 @@
...
+  after a previous votes failed to achieve Sam's result. Sam is
+  attempting to bully other members of core into agreeing, and is thus
+  violating the code of conduct.
+
+Voting items are labeled as either **standard** or **sensitive**.
+Standard items are ones where public record and discourse and record is
+preferable. Sensitive voting items are ones where the results of the
+vote should remain private to the voters after the vote has occurred.
+Sensitive votes should take place on
+`the Helios voting system <https://vote.heliosvoting.org/>`_ in order
```
CJ-Wright: What should we use for standard voting? Can we use Helios for everything?

scopatz: I figured we wouldn't want to, if we just wanted to have a vote durring a meeting

scopatz: It is good tech, though, and other votes can certianly choose to use it.

mariusvniekerk: +1 's on an issue on gh (Apache-ish style?), maybe

mariusvniekerk: Do also want to add something to deal with a minimal number of votes?

CJ-Wright:
Do we need something for a quorum for core meetings?

ocefpaf:
> Do we need something for a quorum for core meetings?

For the meetings I don't think so. Some discussions may be 1-1 sometimes.
For voting we probably should have a minimum required.

scopatz:
> Do we need something for a quorum for core meetings?

I don't think so, because we should be encoraging meetings, even small ones.  

> For voting we probably should have a minimum required

Yeah, so my voting numbers here assumed over *all* conda-forge members.  So implict in this if there are not enough votes cast to make an item pass, it fails.  Maybe that could be more explicit.

src/governance.rst
```diff
@@ -0,0 +1,172 @@
...
+    * Standard
+    * 50% Majority to pass
+
+:Sub-team Dissolution: Proposers must specify the name and justification
+    for why a sub-team should be dissolved.
+
+    * Standard
+    * 50% Majority to pass
+
+:Lock an Issue, Pull Request, Thread: Occasionally, discussions become
```
dougalsutherland: Do we really want to require a vote with a minimum 24-hour voting period and needing at least 50% of core to vote for locking a thread? Seems like a slow process for things that maybe should happen faster.

ocefpaf: I agree. Locking a thread should be "ask for forgiveness and not permission" kind of thing.

scopatz: I have seen this backfire, when threads are closed too quickly. I am not sure I care strongly one way or the other though.  Would you mind PRing in some language to that effect

ocefpaf: I guess that the question is what is worse? Not acting on something that should be locked quickly or opening it again after a wrongful lock?

src/governance.rst
```diff
@@ -0,0 +1,172 @@
...
+  Each member of core is entitled to one vote on all elected matters.
+* **staged-recipes:** The staged-recipes team administers the staged-recipes
+  repository and is responsible for the review and creation of new feedstocks.
+* **maintainers:** A maintainer is an individual responsible for the management
+  of one or more feedstock repositories and packages. Maintainers have the ability
+  to merge pull requests into the feedstock of only the packages they maintain.
+* **external contributors:** This group encompasses all others who are not on
+  core, part of staged-recipes, or maintainers. This includes first time
+  contributors, collaborators, and funders. They have no special rights within
+  the conda-forge organization itself.
```
jakirkham: We also have `conda-forge-webservices`, who maintains the web services (e.g. linter, command services, team updates, feedstock updates, status page, etc.).

src/governance.rst
```diff
@@ -0,0 +1,172 @@
...
+Here are defined the primary teams participating in conda-forge activities.
+
+* **core:** The core team is the governing body over the entire conda-forge
+  organization. Members of the core team have full rights over all conda-forge
+  repositories. Members of core are the face of the project, and are responsible
+  for officially interfacing with external communities, organizations, non-profits,
+  and companies. The core team may create new sub-teams, as appropriate.
+  Each member of core is entitled to one vote on all elected matters.
+* **staged-recipes:** The staged-recipes team administers the staged-recipes
+  repository and is responsible for the review and creation of new feedstocks.
```
jakirkham: Should we include the helper teams?

cc @CJ-Wright

src/governance.rst
```diff
@@ -0,0 +1,172 @@
...
+  core, part of staged-recipes, or maintainers. This includes first time
+  contributors, collaborators, and funders. They have no special rights within
+  the conda-forge organization itself.
+
+Sub-Teams
+---------
+The core team may elect to create new sub-teams for managing the daily business
+of the organization. While sub-teams may have non-core members, every sub-team
+must have at least one core team member at all times. If a sub-team fails to
+have  a core team member for more than 1 week, that team is considered to be
```
jakirkham: nit: Appears there is an extra space here.

src/governance.rst
CJ-Wright: Maybe? Current these are all subteams of staged.

src/governance.rst
```diff
@@ -0,0 +1,172 @@
...
+dissolved. A new sub-team would need to be established by the core in order to
+reinstate the activity.
+
+Sub-teams have a charter that is either *dynamic* or *static*.
+
+* A *dynamic* charter means that the sub-team is self-organizing, with respect
+  to its own internal policies, procedures, and membership. A sub-team may choose
+  to modify its membership independent of the core team. For example, a
+  Google Summer of Code team could be a good candidate for a dynamic charter.
+  Alternatively, language-based maintenance teams also have a dynamic charter.
```
jakirkham: Any thoughts on core's role with dynamic teams? Are they like the team's BDFL or something else?

src/governance.rst
```diff
@@ -0,0 +1,172 @@
...
+  changes must be approved by the core team. For example, a finance team
+  may require a static charter.
+
+All sub-teams must adhere to the governance, policies, and procedures of
+conda-forge at all times.
+
+Voting
+------
+This section presents descriptions and criteria for voting items in the
+conda-forge community. The core team is the only team with voting rights.
```
jakirkham: Are there circumstances when we might want others to have a right to vote as well? As one example that has come up before, should the staged-recipes team be involved in election of their own members?

src/governance.rst
```diff
@@ -0,0 +1,172 @@
...
+retain anonymity.
+
+The default voting period is 1 week (7 days). This may be modified at
+the time when a vote is called, but may never be less than 24 hrs.
+
+:Nominate new member of staged-recipes: The proposer must provide
+    a brief justification as to why the new member is desirable or needed.
+
+    * Sensitive
+    * 50% Majority to pass
```
jakirkham: More [details above]( https://github.com/conda-forge/conda-forge.github.io/pull/598/files#r194808192 ), should staged-recipes be involved in this vote?

src/governance.rst
jakirkham: cc @dougalsutherland

src/governance.rst
jakirkham: What about migration of discussions? IOW the discussion is good and healthy just targeting the wrong forum and we would like to migrate it to an appropriate forum without forking the discussion.

src/governance.rst
```diff
@@ -0,0 +1,172 @@
...
+    members, and charter (dynamic or static) of any new sub-teams.
+
+    * Standard
+    * 50% Majority to pass
+
+:Sub-team Dissolution: Proposers must specify the name and justification
+    for why a sub-team should be dissolved.
+
+    * Standard
+    * 50% Majority to pass
```
jakirkham: If there are no `core` members on the team, doesn't it dissolve anyways? How do we handle this situation?

src/governance.rst
```diff
@@ -0,0 +1,172 @@
...
+
+    * Sensitive
+    * 66.7% Majority to pass
+
+:Remove member of core: The proposer must provide
+    an overwhelming justification as to why the member core
+    should be removed.
+
+    * Sensitive
+    * 75% Majority to pass
```
jakirkham: Commenting here, but this impacts removals and blocks above as well.

It seems like warnings should be raised to the individual or individuals before we come to blocking/removing them. Suggesting this as the individual may not be aware there actions are not inline with the org's expectations (written or otherwise).

How do we do this while still being sensitive as well? For example, mentioning we might block people in a public sphere is likely to have a chilling effect on the community; so, what private communications are appropriate?

How we decide to warn an individual? Does that need a quorum? Potentially a smaller one or specific team dedicated to this function? What determines the make up of this team? How do other orgs handle this?

How many warnings are appropriate? Does this vary depending on the action and/or the member's standing in the org.

Are there forms of behavior that are so egregious that they may deserve fewer or no warnings?

Sorry for the list of questions. Just want to make sure we are giving this a reasonable level of thought.

jakirkham:
Thanks for writing this up, @scopatz. Really appreciate the effort here. Have some questions above. Would be curious to hear yours and others thoughts on these.

src/governance.rst
```diff
@@ -0,0 +1,172 @@
...
+community. This document recognizes that, while packaging is deeply important
+to all of us, conda-forge is an all-volunteer organization. Members of the
+core, any team, or external contributors choose to participate and may
+choose to leave at any time for any reason or for no reason. We deeply
+appreciate all good faith contributions.
+
+Code of Conduct
+---------------
+Conda-forge adheres to the
+`Numfocus Code of Conduct <https://www.numfocus.org/code-of-conduct>`_.
```
jakirkham: Should we link to the versioned copy of the Code of Conduct or note the version?

ref: https://www.contributor-covenant.org/version/1/2/0/code-of-conduct.html

src/governance.rst
ericdill: A lack of "no" on the conda-forge dev call today resulted in an implicit "yes" to helios from everyone on the call @CJ-Wright @jakirkham @mariusvniekerk @jjhelmus @msarahan @ocefpaf @scopatz and "justin"

src/governance.rst
jakirkham: From our discussion earlier, it sounds like this core member exists to form a link between core and the subteam for communication purposes. They may not necessarily even lead the team.

src/governance.rst
jakirkham: IIUC we decided not to explicitly list these

src/governance.rst
```diff
@@ -0,0 +1,173 @@
...
+
+:Remove member of core: The proposer must provide
+    an overwhelming justification as to why the member core
+    should be removed.
+
+    * Sensitive
+    * 75% Majority to pass
+
+All other voting items are considered to be standard and require a 50%
+majority to pass.
```
jakirkham: As discussed in our meeting earlier, we may want a provision for amending this document (as it is a living document). It sounds like the text here mostly matches what we want. We could add a note about amending the document here or perhaps in a nearby section.

src/governance.rst
jakirkham: It's worth noting there is a newer copy of this Code of Conduct, which we should compare the merits of.

ref: https://www.contributor-covenant.org/version/1/4/code-of-conduct

ocefpaf:
@scopatz I guess we should merge this and add additional change in new PRs, are you OK with that?

src/governance.rst
```diff
@@ -0,0 +1,173 @@
+Governance
+==========
+This document outlines the policies and procedures that manage the conda-forge
+community. This document recognizes that, while packaging is deeply important
```
ChrisBarker-NOAA: Are we managing a "community", or a "project" -- I would think project is a better term.

ChrisBarker-NOAA: or maybe "organization"

src/governance.rst
ChrisBarker-NOAA: I think the "latest version" should be the default -- I presume that that link goes to the latest version. If NumFocus happens to update it in a way we don't agree with at some arbitrary point in the future, we can revisit then.

src/governance.rst
```diff
@@ -0,0 +1,173 @@
...
+Standard items are ones where public record and discourse and record is
+preferable. Sensitive voting items are ones where the results of the
+vote should remain private to the voters after the vote has occurred.
+Sensitive votes should take place on
+`the Helios voting system <https://vote.heliosvoting.org/>`_ in order
+retain anonymity.
+
+The default voting period is 1 week (7 days). This may be modified at
+the time when a vote is called, but may never be less than 24 hrs.
+
```
ChrisBarker-NOAA: In the following the term "majority" is used -- is that a majority of the entire core team? or a majority of people that placed a vote on the particular issue?

If the former, then it may be very hard to attain a majority quickly when needed. If the later, than a minimum quorum should be defined.

In any case, we need a bit of clarifying language.

src/governance.rst
```diff
@@ -0,0 +1,173 @@
...
+    * Sensitive
+    * 60% Majority to pass
+
+
+:Remove member of staged-recipes: The proposer must provide
+    a justification as to why the member of staged recipes
+    should be removed.
+
+    * Sensitive
+    * 66.7% Majority to pass
```
ChrisBarker-NOAA: Do we really need some things with 60%, and some with 66.7% ?

Maybe 1/2, 2/3, 3/4 as options is enough? (or only 1/2 and 2/3?)

ChrisBarker-NOAA: Note that there are currently 15 members, so 2/3 is well defined :-)

scopatz:
> I guess we should merge this and add additional change in new PRs, are you OK with that?

@ocefpaf, yes! totally OK with that

ocefpaf:
Merging. Most of the comments here should be PRs and hopefully the person making the comment will send that PR :wink:

scopatz:
Thanks @ocefpaf! Yes, please @ChrisBarker-NOAA @jakirkham et al. put in additional PRs.  I am happy to merge those!

PR #602, opened 2018-06-19
Adds new branch reference to webpage and docs

### LINKED ISSUE #600: DOC: Specify using non-master branch on webpage and docs
Recently sat with a coworker on their first PR and noticed the webpage doesn't suggest to use a different branch other than `master` for a PR. Would be good to have this noted on the webpage and in the docs.

jrbourbeau:
This PR adds a note about using a non-master branch to the webpage and docs. 

Closes #600.

### DIFF
--- a/src/recipe.rst
+++ b/src/recipe.rst
@@ -32,12 +32,13 @@ In case you are building your first recipe using conda-forge, a step-by-step ins
 Step-by-step Instructions
 -------------------------
 
-1. Fork `example recipes <https://github.com/conda-forge/staged-recipes/tree/master/recipes>`_
-2. Within your forked copy, generate a new folder in the recipes subdirectory and copy the `meta.yml <https://github.com/conda-forge/staged-recipes/blob/master/recipes/example/meta.yaml>`_ file from the example directory. Please leave the example directory unchanged!
-3. Edit the copied recipe (meta.yml) as needed.
-4. Generate the sha-256 key as described in the example recipe using the ``openssl`` tool. As an alternative you can also go to the package description on `PyPi <https://pypi.org>`_ from which you can directly copy the sha-256.
-5. Be sure to fill in the ``tests`` section. The simplest test will simply test that the module can be imported, as described in the example.
-6. Remove all irrelevant comments in the ``meta.yaml``  file.
+1. Fork the `example recipes <https://github.com/conda-forge/staged-recipes/tree/master/recipes>`_ repository.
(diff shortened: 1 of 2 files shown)


jrbourbeau:
@jakirkham let me know if I missed another place this should added to in the docs. Or if a rewording is needed.

src/recipe.rst
```diff
@@ -32,12 +32,13 @@ In case you are building your first recipe using conda-forge, a step-by-step ins
...
-4. Generate the sha-256 key as described in the example recipe using the ``openssl`` tool. As an alternative you can also go to the package description on `PyPi <https://pypi.org>`_ from which you can directly copy the sha-256.
-5. Be sure to fill in the ``tests`` section. The simplest test will simply test that the module can be imported, as described in the example.
-6. Remove all irrelevant comments in the ``meta.yaml``  file.
+1. Fork the `example recipes <https://github.com/conda-forge/staged-recipes/tree/master/recipes>`_ repository.
+2. Create a new branch from the staged-recipes ``master`` branch.
+3. Within your forked copy, generate a new folder in the recipes subdirectory and copy the `meta.yml <https://github.com/conda-forge/staged-recipes/blob/master/recipes/example/meta.yaml>`_ file from the example directory. Please leave the example directory unchanged!
+4. Edit the copied recipe (meta.yml) as needed.
+5. Generate the sha-256 key as described in the example recipe using the ``openssl`` tool. As an alternative you can also go to the package description on `PyPi <https://pypi.org>`_ from which you can directly copy the sha-256.
```
scopatz: It is usually spelled SHA256

scopatz: Lines should be limited to 80 chars.

scopatz: Also using the sha256sum command is much easier than opens in my opinion.

jrbourbeau: Thank you for the review @scopatz, I'll update accordingly 

> Also using the sha256sum command is much easier than opens in my opinion.

Agreed re: `openssl`, but I think that can be a separate PR to update the recommended way to get the SHA256 key.  It will also need to be updated in the example `meta.yml` in `staged-recipes`, which currently recommends using `openssl` 

https://github.com/conda-forge/staged-recipes/blob/73d7fe0d2695a605ef4adb3f390c77ea1f07806d/recipes/example/meta.yaml#L18-L22

jakirkham:
Looks great! Especially appreciate that you took the time to find the neat branching icon. Thanks for doing this @jrbourbeau. :)

PR #609, opened 2018-07-19
staged-recipes is not mandatory for core members

ocefpaf:
@scopatz can you review this one?

### DIFF
--- a/src/governance.rst
+++ b/src/governance.rst
@@ -88,9 +88,8 @@ the time when a vote is called, but may never be less than 24 hrs.
 
 :Nominate new member of core: The proposer must provide
     a sufficient justification as to why the nominee should be welcomed
-    into core. It is necessary that all nominees are current and active
-    members of staged-recipes. Nominees to core must have been members
-    of staged-recipes for at least 1 month (4 weeks) prior to this vote.
+    into core. It is not necessary that all nominees are active
+    members of staged-recipes but that is desirable.
 
     * Sensitive
     * 66.7% Majority to pass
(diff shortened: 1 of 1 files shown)


src/governance.rst
```diff
@@ -88,9 +88,8 @@ the time when a vote is called, but may never be less than 24 hrs.
 
 :Nominate new member of core: The proposer must provide
     a sufficient justification as to why the nominee should be welcomed
-    into core. It is necessary that all nominees are current and active
-    members of staged-recipes. Nominees to core must have been members
-    of staged-recipes for at least 1 month (4 weeks) prior to this vote.
+    into core. It is not necessary that all nominees are active
```
CJ-Wright: Maybe we could state this more generically? 

eg
`Prior service to the community, including but not limited to: serving as a staged-recipes reviewer, working on critical conda-forge infrastructure, and helping to bridge disparate communities are an important part of the nomination process`.

scopatz:
Thanks for putting this in @ocefpaf! I like @CJ-Wright's wording a bit better.

src/governance.rst
ocefpaf: I guess that "a sufficient justification a" is better than listing all possible reasons to justify the invitation and then modifying it when we find an issue, like this one.

ocefpaf:
> I like @CJ-Wright's wording a bit better.

I don't b/c I'm concerned that people will focus on that instead of using common sense. Hopefully the "including but not limited to" will prevent that from happening.

scopatz:
Thanks @ocefpaf @CJ-Wright!

PR #623, opened 2018-08-07
add lori

ocefpaf:
@loriab can you please review if this is correct?


scopatz:
@loriab should also be added to the end of the governance document too :wink:

ocefpaf:
> @loriab should also be added to the end of the governance document too wink

Thanks. I'll do this here to avoid another PR. (But we may try to automate this so we edit only one file :smile:)

scopatz:
Yeah that makes sense, we could easily add something to conf.py that generates this listing from the CSV file.

loriab:
Sorry, just catching up. All correct, thanks.

PR #624, opened 2018-08-08
add a disclaimer about update-cb3

dougalsutherland:
Just adds a warning to the docs about the `update-cb3` webservices command to be careful with the output.


ocefpaf:
I should've done this a long time ago! Thanks @dougalsutherland!!

PR #642, opened 2018-09-04
added donate button

scopatz:
Now that we have a mechanism to receive donations :)

### DIFF
--- a/index.html
+++ b/index.html
@@ -66,6 +66,9 @@
                     <li>
                         <a href="status">Status</a>
                     </li>
+                    <li>
+                        <a href="shttps://www.flipcause.com/secure/cause_pdetails/NDA0OTk=">Donate!</a>
+                    </li>
                 </ul>
             </div>
             <!-- /.navbar-collapse -->


ocefpaf:
We should probably add that to conda-smithy to add that in all feedstocks after a re-render. What do you think?

jakirkham:
Having a button somewhere prominent seems nice. Having it everywhere seems a bit intrusive IMHO.

It would be good if we could explain what we do with donations (that of course assumes we know what we do with donations 😄).

scopatz:
> We should probably add that to conda-smithy to add that in all feedstocks after a re-render

I agree that this is something we should do. Chatting with @mariusvniekerk today, he indicated that he thought this was a bit aggressive, like @jakirkham implies :smile:.

I think that if we got donations to the $2 - 3K/year level, we'd have enough to start securing  CI services. This would help us perform migrations faster, shorten queue times etc.

Maybe we should take a vote on this

ocefpaf:
I don't think it is aggressive and I think a lot of people interact with conda-forge via 1-2 feedstocks only. Maybe instead of a donate button we could add a link about conda-forge itself that points to the main pages.m (Our main page is a desert actually, very few people visit it. The feedstocks on the other hand a visited frequently!)

scopatz:
Well, in any event, this is not a PR to add the donate button everywhere, just to the main page.

jakirkham:
Agreed. Thanks @scopatz.

We should probably figure out how we want to budget/allocate money that is given to us and share that information with the community. Would also be good if we can figure out some procedure for how we change the budget (e.g. proposing, time for public comment, voting, etc.).

marcelotrevisani:
Nice!!

PR #669, opened 2018-10-31
meta.yaml: cross-ref detailed docs

ax3l:
I found it hard as a beginner to find the right and up-to-date docs for writing a `meta.yaml` from scratch. This PR tries to improve this by linking to the more detailed `conda` docs immediately instead of little side-note.


ax3l:
wow, that was fast :smiley:

PR #670, opened 2018-11-03
describe how to solve the openblas/mkl dance.

hmaarrfk:
https://github.com/conda-forge/numpy-feedstock/issues/108

With this we can probably close the issue above.

cc: @ocefpaf

### DIFF
--- a/src/blas.rst
+++ b/src/blas.rst
@@ -66,3 +66,57 @@ BLAS & Numpy & Friends
 
    -  Same thing as NumPy
    -  Add numpy dependency as if linking occurs
+
+openblas mkl dance
+------------------
+
+When updating packages, it might seem that openblas and mkl keep trying to
+overwrite one and other. For example:
+
+.. code-block:: bash
+
+  $ conda install pytest
+  Solving environment: done
+
+  [...]
+
+  The following packages will be UPDATED:
+
+  libgcc-ng:      7.2.0-hdf63c60_3                     conda-forge --> 8.2.0-hdf63c60_1
+  numpy:          1.15.2-py36_blas_openblashd3ea46f_1  conda-forge [blas_openblas] --> 1.15.2-py36h1d66e8a_1
+
+  The following packages will be DOWNGRADED:
+
+  blas:           1.1-openblas                         conda-forge --> 1.0-mkl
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
It could be interesting to adopt the `nomkl` approach of default with the mutex package.

hmaarrfk:
@SylvainCorlay can you explain a little more. .This seems beyond my understanding of numpy.

src/blas.rst
```diff
@@ -66,3 +66,57 @@ BLAS & Numpy & Friends
...
+in the environment. Implicit dependencies, such as openblas in the case of
+``numpy`` from conda-forge, behave differently from explicit ones.
+Explicitly specifying the dependency on either ``openblas`` or ``mkl`` will
+solve this problem. As of writing, conda-forge does not package ``mkl``.
+
+Specifying:
+
+.. code-block:: bash
+
+  conda install blas=*=openblas
```
croth1: Most unix shells require quotation around the star `conda install "blas=*=openblas"`

hmaarrfk: thanks

src/blas.rst
```diff
@@ -66,3 +66,57 @@ BLAS & Numpy & Friends
...
+in the environment. Implicit dependencies, such as openblas in the case of
+``numpy`` from conda-forge, behave differently from explicit ones.
+Explicitly specifying the dependency on either ``openblas`` or ``mkl`` will
+solve this problem. As of writing, conda-forge does not package ``mkl``.
+
+Specifying:
+
+.. code-block:: bash
+
+  conda install "blas=*=openblas"
```
jakirkham: Just as a note, if users want to force channel priority as well, they can do the following.

`conda install "conda-forge::blas=*=openblas"`.

jakirkham:
Thanks @hmaarrfk 😄

src/blas.rst
hmaarrfk: hmm good point. should we make that change?

mathause:
For me this problem was also resolved by setting

    channel_priority: strict

in my .condarc

hmaarrfk:
Yes it seems that there have been some changes to how the solver works. Not sure what the best strategy is

PR #688, opened 2018-12-21
Add note about local testing config w/ conda-build

ihnorton:


src/recipe.rst
```diff
@@ -66,6 +66,9 @@ Checklist
 * In case your project has tests included, you need to decide if these tests should be executed while building the conda-forge feedstock.
 * Make sure that all tests pass sucessfully at least on your development machine.
 * Recommended: run ``conda-build`` on your source code to ensure the recipe works locally.
+  **In order to build against the correct versions of library dependencies specified in `meta.yaml`, the proper conda-forge channel must be added to your conda config**:
+    - conda config --add channels conda-forge/label/gcc7 # when using the conda3 compiler toolchain
+    - conda config --add channels conda-forge
```
scopatz: Can you write something about how both of these should not be acvtive at the same time? Thanks!

croth1:
I am closing this as it is probably not relevant anymore.

ihnorton:
Aside from gcc7 line, the instructions are still relevant because conda build does not have a mechanism to specify package sources in the metadata.

croth1:
I agree that the information generally is valuable and should end up somewhere in the documentation. However it is not really clear to me where. At the moment information for local testing is spread all over the docs e.g.:
- http://conda-forge.org/docs/maintainer/adding_pkgs.html#running-tests-locally-for-staged-recipes
- http://conda-forge.org/docs/maintainer/knowledge_base.html#local-testing

I think the clean way is to start a new section in the "knowledge base" section and gather all this information in one place.

I generally use the docker build script for local testing. I have no experience with windows or macos testing, so I am probably not the right person to compose this section.

PR #704, opened 2019-02-05
Remove deprecated mention of --no-deps

chrisburr:
Since #653 the recommended `pip` installation command hasn't included `--no-deps`. This PR edits the related text to be consistent.

### DIFF
--- a/src/meta.rst
+++ b/src/meta.rst
@@ -371,8 +371,7 @@ These options should be used to ensure a clean installation of the package witho
 dependencies. This helps make sure that we're only including this package,
 and not accidentally bringing any dependencies along into the conda package.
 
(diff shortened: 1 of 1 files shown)


scopatz:
There actually should be a no-deps

chrisburr:
I thought [`--no-deps` was now set by `conda-build`](https://github.com/conda-forge/conda-forge.github.io/pull/653#discussion_r223453603)?

If not, I'll make a new PR. Should `--ignore-installed` also be included again?

scopatz:
I think at this point only the `--no-deps` is needed. The line that is in the staged-recipes example is correct. The docs here are out of date

PR #705, opened 2019-02-05
Recommended using --no-deps with pip

chrisburr:
Replaces #704 and adds `--no-deps` to the recommended pip install command.


scopatz:
Thanks @chrisburr!

PR #708, opened 2019-02-06
Search

scopatz:
This implements a basic front end to the libcflib.rest.search endpoint.  It could obviously use some beautification, but it is an MVP in its present form. 

This is implemented in elm and I added some tools for building and integrating with our CI.


scopatz:
CC @conda-forge/core

scopatz:
Pinging this!

PR #709, opened 2019-02-06
meta.yml, omit fn key

marscher:
If the url specifies the filename (no weird redirection etc.), we can safely omit the file name.


scopatz:
Thanks!

Issue #645, opened 2018-09-10
Feedback as requested by @jakirkham on Gitter


PhilipMay:
Hi everybody,
my name is Philip and I am new to conda-force.
At the moment I am maintaining the hyperopt package.
@jakirkham asked me on Gitter to give some feedback.
I think this is a very good idea - so here it is:

1. When I was first thinking to build a conda package
I had a look into conda-forge and asked myself: "Hey,
what's the difference of uploading packaged to anaconda.org
or publishing them to conda-forge?" 
Maybe you could publish a chapter where you talk about 
the differences and advantages.

2. The "Packaging the License Manually" chaper is not 100% clear to me.
For me it ~~was~~ is not 100% clear if the LICENSE.txt file must be included
in the same directory as the meta.yaml or not. Even when it is included
in the tarball.
See here: https://conda-forge.org/docs/meta.html#packaging-the-license-manually

3. Everything else was much to read but good and understandable to me. 

Thanks for your work
Philip

scopatz:
Hi @PhilipMay - thanks for the feedback!  Please feel free to put in PRs that make the text of the documentation more clear to you. All of those docs are in this repo :)

PhilipMay:
@scopatz yes - well, good idea. But actually I do not know the exact answer for both topics. So it would be cool if someone else could do that.

scopatz:
@PhilipMay - from what you said, it seems like you understand (2) :wink:

PhilipMay:
> @PhilipMay - from what you said, it seems like you understand (2) 😉

Well, yes you are right. I changed the text. Sorry for the confusion.

croth1:
In the linked commit introduced changes that hopefully clarify the docs. Thanks for your feedback @PhilipMay!

Issue #607, opened 2018-07-15
Clarifying docs


ChayaSt:
@msarahan, here are some points that can be clarified in the documentation: 

1. There is no explanation why conda-forge (vs default conda channel) in the documentation. This should probably be explained in the beginning of the about page. 
2.  The section below checklist is titled "What happens after the PR to staged-recipes is merged". This can be a little confusing. It might be clearer to add a step to merge the PR to staged recipes after going through the checklist in the step by step instructions https://conda-forge.org/docs/recipe.html#step-by-step-instructions

scopatz:
Thanks for brining this up @ChayaSt! Pull requests are definitely welcome!

croth1:
I made following changes to the docs:
 http://conda-forge.org/docs/user/introduction.html#why-conda-forge addressing (1)
 http://conda-forge.org/docs/maintainer/adding_pkgs.html#feedback-and-revision addressing (2)

Thanks @ChayaSt for giving feedback!

PR #728, opened 2019-03-07
A handful of doc improvements

croth1:
- Moved introduction to user section
- addressed feedback from issues (#607, #645)


scopatz:
Thanks a million @croth1!

PR #731, opened 2019-03-08
clarify docs on noarch and py3-only packages

ngoldbaum:
The constraint needs to be in both the host and run sections so I've made that clearer.


scopatz:
Thanks @ngoldbaum!

PR #740, opened 2019-03-17
Handful of small doc improvments

croth1:
a couple of small changes

 - added a clickable (Q) to every FAQ question that serves as a means to get a link with an anchor that points directly to the question
 - optimize the style of the faq to make it even more readable
 - initialised a faq section for maintainers
 - linked two issues to make it easier to find related discussion


scopatz:
Thanks!

PR #766, opened 2019-04-05
add a strict note about the compilers run-time

ocefpaf:
Ping @xylar

### DIFF
--- a/src/user/tipsandtricks.rst
+++ b/src/user/tipsandtricks.rst
@@ -35,7 +35,7 @@ Note: All of conda-forge software pinning can be found at: https://github.com/co
 How to fix it?
 --------------
 
-Newer ``conda`` versions introduced a strict channel priority feature.
+Newer ``conda`` versions (>=4.6) introduced a strict channel priority feature.
 Type ``conda config --describe channel_priority`` for more information.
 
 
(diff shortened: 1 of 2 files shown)


src/user/tipsandtricks.rst
```diff
@@ -61,6 +61,15 @@ Here is how a ``.condarc`` file would look like:
 In addition to the channel priority we recommend to always install your packages inside a new environment instead the root environment from anaconda/miniconda.
 Using envs make it easier to debug problems with packages and ensure the stability of your root env.
 
+Note: in the past ``conda-forge`` used to vendorize some of ``defaults`` dependencies that were not built in our infrastructure,
```
croth1: Can we put this in a `..note::` box for consistency?

src/user/tipsandtricks.rst
```diff
@@ -61,6 +61,15 @@ Here is how a ``.condarc`` file would look like:
...
+like compilers run-times, to avoid the mixing channel problem.
+However, with the ``strict`` option, we no longer have to vendorize those (which has its own problems too),
```
croth1: Do we now recommend everybody to use strict? If so we probably should also put that here: http://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge

croth1: I personally would leave out `(which has its own problems too)` but I guess that's a matter of taste.

xylar: Yes, I agree.

xylar:
```suggestion
However, with the ``strict`` option, we no longer have to vendorize those (this led to its own set of problems),
```
@croth1 it seems like a good idea to indicate why these channels are marked as `broken` and effectively not available.  My preference would be for some brief explanation of what those problems where but it seems good to at least indicate that there was a good reason not to keep them available, given that it's caused some inconvenience for people like me that where told to use only the `conda-forge` channel before the `strict` option was available.

ocefpaf: In `conda 4.7` where `strict` will be default and we will need to update that but I guess that, b/c mixing channels without `strict` is broken, we should add it there too.

src/user/tipsandtricks.rst
ocefpaf: Done. Thanks!

src/user/tipsandtricks.rst
ocefpaf: Done! Not sure if we should leave the comment about conda 4.7 there though.

ocefpaf: Maybe a better explanation for this should be in the `News` section?

ocefpaf:
@croth1 and @xylar this is ready for a second round of review/merge.

src/user/tipsandtricks.rst
xylar: Sure, that's fine.

xylar:
Yep, looks good to me.

croth1:
Thanks @ocefpaf @xylar for working on this!

PR #768, opened 2019-04-08
MAINT: libGL - rerender reminder

andyfaff:
I was trying to get libGL.so to work on https://github.com/conda-forge/refnx-feedstock/pull/20, but having major difficulties. I wasn't aware that a re-render was necessary. This PR points out that as well as making changes to meta.yaml and yum_requirements.txt, that a re-render is necessary.


ocefpaf:
Thanks! We do have that information in the `yum_requirements` section but it makes sense repeating it here for clarity. (And most people searching for the CDT instructions won't follow the link to the `yum_requirements` b/c that is not what they searched in the first place.)

PR #772, opened 2019-04-13
Remove avoiding-the-openblas-mkl-dance

ocefpaf:
Closes https://github.com/conda-forge/conda-forge.github.io/issues/771


ocefpaf:
> Fine with me. I don't think we have a linked that section anywhere do we? If so we may want to change those links before we delete this.

I could not find it but I have to confess I only tried a few `find` keywords combinations at the CLI.  BTW, would sphinx `linkcheck` catch something like this? I know it catches external links, not sure about internal references.

croth1:
Just thought about it again: since we never labelled the header, there should be no references. I think sphinx would warn automatically when you build locally. I think you can merge 👍

ocefpaf:
> I think you can merge +1

You do have merge rights here, right? Just checking in case you need it...

croth1:
@ocefpaf, yes I am maintainer here 👍

PR #780, opened 2019-04-25
doc: faq entry for build numbers >1000 (#779)

croth1:
xref: #779.


ocefpaf:
Awesome! Thanks for this explanation in the docs!!

PR #784, opened 2019-05-10
Change "blockers" to "conflicts"

chrisburr:
Following on from a question on gitter, I think using the word "conflicts" is more discoverable.

I've left "blockers" in the body of the text so searching can find either term.


croth1:
I completely agree. Thanks 👍

PR #793, opened 2019-05-30
update quick-usage instructions

carlodri:
as per http://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge


scopatz:
Thanks!

PR #795, opened 2019-06-03
First draft on emeritus core

ocefpaf:
As we discussed in our past meeting the emeritus-core still have all the rights of a core member but they will not count to the total number of core members to avoid sticky situations with polls that need majority votes to pass.

PR #801, opened 2019-07-03
Initial move of inactive members to emeritus

ericdill:
Hi @patricksnape @pelson and @bgruening,

As per the recent addition to our governance docs (in #795) we have added the notion of an emeritus member of the conda-forge community. The primary way in which your role is changing is called out in #795, but bears repeating here:

> Emeritus core members can still vote and be back to active core any time, the only difference is that emeritus-core will not count as the total core members when computing the necessary votes a poll needs to pass.

The @conda-forge/core team would like to thank you for contributions to the community. We will happily welcome you back to /core if you become active again on the core team!


scopatz:
@ocefpaf - feel free to merge

PR #808, opened 2019-07-12
Clarify how to add a new recipe.

jasongrout:
I was confused about exactly where I should put my recipe, and if I was supposed to copy or rename the example. I clarified both of these points.


jakirkham:
Thanks @jasongrout!

PR #809, opened 2019-07-12
Use square brackets instead of angle brackets in HTML

jasongrout:
In #808, I used angle brackets to denote a name the user should replace. Of course, this did not render in HTML. This replaces the angle brackets with square brackets.


ericdill:
Thanks @jasongrout !

PR #817, opened 2019-07-19
remove search

scopatz:
I took down the service that backended this to save conda-forge some money. Makes sense to remove it from the docs


scopatz:
This seem R4R @conda-forge/core

PR #834, opened 2019-08-06
fix my emeritus status

CJ-Wright:


CJ-Wright:
I forgot to remove myself from the core.csv and GH balked at my double quotes (black has trained me well)

PR #840, opened 2019-08-18
improve how to use docs

ocefpaf:
This was based on the feedback from the SciPy sprint.

### DIFF
--- a/src/user/introduction.rst
+++ b/src/user/introduction.rst
@@ -58,6 +58,13 @@ and activate the ``strict`` channel priority (strict will be activated by defaul
 
 From now on using ``conda install <package-name>`` will also find packages in our conda-forge channels.
 
+.. note::
+
+  In addition to the channel priority,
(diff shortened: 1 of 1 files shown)


src/user/introduction.rst
```diff
@@ -58,6 +58,13 @@ and activate the ``strict`` channel priority (strict will be activated by defaul
 
 From now on using ``conda install <package-name>`` will also find packages in our conda-forge channels.
 
+.. note::
+
+  In addition to the channel priority,
```
ocefpaf: There is some redundancy with the docs in the "tips and trick" section but I believe we need to:

- keep the "tips and trick" as-is b/c there are lots of issues that points to it;
- add it here b/c it is a simple "how to use" questions that we never answer properly.

99% percents of the problems I have to solve about conda-forge is the use of envs and avoid installation in the `base` env.

PR #841, opened 2019-08-19
Entry points can also be defined in `setup.cfg`

nsoranzo:
E.g. https://github.com/pypa/virtualenv/blob/master/setup.cfg#L61

Checklist
* [ ] Used a fork of the feedstock to propose changes
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

### DIFF
--- a/src/maintainer/knowledge_base.rst
+++ b/src/maintainer/knowledge_base.rst
@@ -689,7 +689,7 @@ In order to qualify as a noarch python package, all of the following criteria mu
     section.
   - ``2to3`` is not used
   - Scripts argument in setup.py is not used
-  - If ``console_script`` entry points are defined in ``setup.py`` or ``setup.cfg``, they are also listed in the ``build`` section of ``meta.yaml``
+  - If ``console_script`` ``entry_points`` are defined in ``setup.py`` or ``setup.cfg``, they are also listed in the ``build`` section of ``meta.yaml``
   - No activate scripts
   - Not a dependency of ``conda``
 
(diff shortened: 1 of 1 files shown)


ocefpaf:
/rant 
I really miss the "there is only one obvious way" philosophy in Python. The packaging world is becoming the real life version of the XKCD standards cartoon.
rant/

src/maintainer/knowledge_base.rst
```diff
@@ -687,11 +687,11 @@ In order to qualify as a noarch python package, all of the following criteria mu
   - No skips except for python version. If the recipe is py3 only, remove skip
     statement and add version constraint on python in ``host`` and ``run``
     section.
-  - 2to3 is not used
+  - ``2to3`` is not used
   - Scripts argument in setup.py is not used
-  - If ``console_script`` entrypoints are in setup.py, they are listed in meta.yaml
+  - If ``console_script`` entry points are defined in ``setup.py`` or ``setup.cfg``, they are also listed in the ``build`` section of ``meta.yaml``
```
ocefpaf: I would keep it one word b/c that is how people will find it in the `setup.*` files.

nsoranzo: The world is actually `entry_points`, which I had used at a certain point while editing this, but then I have to also quote it as code right after another code-quoted word. But I can do that if you prefer.

ocefpaf: Indeed. I think it is better.

PR #848, opened 2019-08-30
Update note strict channel priority in conda 5.0

lucatrv:
See: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html#strict-channel-priority


ocefpaf:
Thanks @lucatrv!

PR #872, opened 2019-09-24
min conda version

ocefpaf:
Pretty much everybody is already on latest `conda` but I think it is nice to say that explicitly in the docs.


scopatz:
Thanks!

Issue #852, opened 2019-09-04
Easier path for "how to get help" on the website


djsutherland:
Where/how to submit an issue is not always obvious to users who don't really know about how conda-forge is organized; @ocefpaf reported that many users would rather just work around stuff themselves instead rather than reporting issues upstream.

It might help to have a big "help" button (or similar) on the website, and maybe linked from various repos as well, describing where to go to report issues: how to find a feedstock repo for a given package name, what kinds of things go on the webpage repo, etc.

We could also think about ways to make this more friendly, especially in the case when feedstock maintainers don't respond right away.

viniciusdc:
@ocefpaf Any suggestions on what we should address to this besides the #880 ? (If not, we can close this or reformulate as an Outreachy issue)

ocefpaf:
I'm OK closing this. If @djsutherland feels that #880 is not really what she thought on when opening this issue we can re-open it later.

PR #880, opened 2019-10-01
how to get help

ocefpaf:
Partially addresses https://github.com/conda-forge/conda-forge.github.io/issues/852

I don't really like what I wrote here and there are lots of room for improvements! Please, if possible, push suggestions and changes directly to this PR.

### DIFF
--- a/src/user/00_intro.rst
+++ b/src/user/00_intro.rst
@@ -9,4 +9,4 @@ User Documentation
   tipsandtricks
   faq
   contributing
-  
+  how_to_get_help
--- a/src/user/how_to_get_help.rst
+++ b/src/user/how_to_get_help.rst
@@ -1,0 +1,39 @@
+How to get help in conda-forge
+==============================
+
+We have three options for communication,
+live `chat on gitter <https://gitter.im/conda-forge/conda-forge.github.io>`__,
+GitHub issues,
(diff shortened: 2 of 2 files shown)


ocefpaf:
> Where do you see conda-forge.github.io fitting into this?

Good question! I could say I did not think about it but the truth is that at the back of my brain was just hoping to see it replaced by the mailing list :smile:

What do you suggest?

CJ-Wright:
That works for me, the mailing list is not used that much at the moment.

ocefpaf:
> the mailing list is not used that much at the moment.

Yeah. Not sure if that is b/c people are moving away from mailing lists or b/c we never advertised it properly.

CJ-Wright:
Personally I have a slight preference for the GH issues, since they are easier to track (closed v open, labels, assignments, etc.)

ocefpaf:
> Personally I have a slight preference for the GH issues, since they are easier to track (closed v open, labels, assignments, etc.)

Should we add `conda-forge.github.io` as a generic-long-threads thing then and place the mailing list as an less preferred alternative to that?

CJ-Wright:
@conda-forge/core thoughts? I think this is ready for review.

scopatz:
LGTM! Thanks!

ocefpaf:
Thanks @isuruf!

PR #914, opened 2019-10-26
docs: reverse minutes order - most recent one now on the top

croth1:
Reversing the minutes so that the most recent meeting is on the top:

image


scopatz:
Thanks!

PR #925, opened 2019-11-19
add steps to complete migrations

minrk:
issue PR to update pinned version and remove the migrator

This is also for my own edification, since I'm not actually 100% sure that this is how migrations are supposed to be completed.

Added:

- migration status can be checked at conda-forge.org/status
- once complete, PR to update pinning, pin package date, and remove migrator yaml

Semi-related: Should the bot issue this PR itself when it notices that the migration is complete?


scopatz:
Thanks @minrk!

PR #948, opened 2020-01-02
docs: describe definition of `CONFIG` variable

Midnighter:
Fix https://github.com/conda-forge/staged-recipes/issues/9961

When testing a build locally, not having `CONFIG` defined led to an error. So I think it should be part of the docs.


scopatz:
Thanks @Midnighter!

PR #960, opened 2020-01-27
Add FAQ entry for conda package not appearing

saulshanabrook:
I was confused when I released a new version of a package, and it showed up on the package page, but I wasn't able to download it locally. @scopatz explained to me this is because of a CDN. I added this explanation to the FAQ to help others who have the same question.


scopatz:
Thanks @saulshanabrook!

PR #970, opened 2020-02-04
clarify rules around quorum, add timeout proposal

msarahan:
@conda-forge/core
This PR falls under the Modifying Governance policy, please vote and/or comment on this PR.
This PR needs 75% of core to vote yea to pass.
To vote please leave Approve (yea) or Request Changes (nay) reviews.
If you would like changes to the current language please leave a comment or push to this branch.
This vote will end on February 19, after the general meeting on that day.

### DIFF
--- a/src/orga/governance.rst
+++ b/src/orga/governance.rst
@@ -109,14 +109,15 @@ To call for a standard vote, here is a template PR comment:
 :Quorum: All percentages below express *both* required participation, as a
          fraction of the active core team, as well as the fraction of that
          fraction who vote affirmatively on the issue. For example, in a vote
-         requiring 50%, with 18 active core members, at least 9 must vote, and
-         among them, there must 5 affirmative votes.
+         requiring 50%, with 18 active core members, at least 9 must vote;
+         if 9 vote, there must 5 affirmative votes. If 13 members vote, 7
+         must be affirmative.
 
          Because everyone is busy and it is difficult sometimes to get quorum,
-         there is a timeout, where the current participation level is taken for
-         what it is, and the percentage of affirmative votes is calculated from
-         whatever the vote total is at that time.  In order for a timeout to occur,
-         the vote must have:
+         votes not achieving quorum will eventually time out. When this happens,
+         the current participation level is taken for what it is, and the percentage
+         of affirmative votes is calculated from whatever the vote total is at that
+         time.  In order for a timeout to occur, the vote must have:
 
              * been open for at least 2 weeks
              * been advertised on at least 3 separate occasions on the gitter core
@@ -139,8 +140,8 @@ To call for a standard vote, here is a template PR comment:
               @conda-forge/core
               This vote falls under the {policy} policy, please vote and/or comment on this PR.
               This vote needs {policy_percent} of core to vote yea to pass.
-              This vote presently has {current voters}, and needs {policy_percent * core} more for quorum.
-              It is proposed that this vote will timeout and be evaluated with the current votes in {days}, on {date}.
+              This vote presently has {current_voters}, and needs {policy_percent * core - current_voters} more for quorum.
+              It is proposed that this vote will time out and be evaluated with the current votes in {days}, on {date}.
               To vote please leave Approve (yea) or Request Changes (nay) reviews.
 
 ----
--- a/src/orga/governance.rst
+++ b/src/orga/governance.rst
@@ -166,7 +166,7 @@ To call for a standard vote, here is a template PR comment:
               It is proposed that this vote will time out and be evaluated with the current votes in {days}, on {date}.
               To vote please leave Approve (yea) or Request Changes (nay) reviews.
               
-         To declare a vote "timed out," the person making such a declaration must post a pull-request adding 
+         To declare a standard vote "timed out," the person making such a declaration must post a pull-request adding
          a vote record to the https://github.com/conda-forge/conda-forge.github.io/tree/master/src/vote-results
          folder.  The declaration PR should be merged by the first core member who is available to verify that
          the requirements for the timeout have been met, based on their own personal records.
--- a/src/orga/governance.rst
+++ b/src/orga/governance.rst
@@ -96,6 +96,8 @@ the authoritative list of emails to use.
 The default voting period is 1 week (7 days). This may be modified at
 the time when a vote is called, but may never be less than 24 hrs.
 
+Additional requirements may apply in case low turnouts have to be handled, see "Quorum" below.
+
 To call for a standard vote, here is a template PR comment:
 
 .. code-block:: md
--- a/src/orga/governance.rst
+++ b/src/orga/governance.rst
@@ -106,6 +106,23 @@ To call for a standard vote, here is a template PR comment:
     To vote please leave Approve (yea) or Request Changes (nay) reviews.
     If you would like changes to the current language please leave a comment or push to this branch.
     This vote will end on {date}.
+    
+----
+:Posting results: To maintain the historical record, the outcome of any vote should be 
+         recorded in the "vote-results" folder at 
+         https://github.com/conda-forge/conda-forge.github.io/tree/master/src/vote-results
+         
+         Each vote should be its own file.  The filename should reflect the topic and the 
+         date that the vote opened.  The file should contain at least:
+         
+         * vote description
+         * vote policy
+         * vote totals
+         * Poll open and close dates
+         
+         If a vote has "timed out," as described in "Quorum," the documentation regarding
+         notification records should also be in this file.
+         
 
 ----
 :Quorum: All percentages below express *both* required participation, as a
@@ -145,6 +162,11 @@ To call for a standard vote, here is a template PR comment:
               This vote presently has {current_voters}, and needs {policy_percent * core - current_voters} more for quorum.
               It is proposed that this vote will time out and be evaluated with the current votes in {days}, on {date}.
               To vote please leave Approve (yea) or Request Changes (nay) reviews.
+              
+         To declare a vote "timed out," the person making such a declaration must post a pull-request adding 
+         a vote record to the https://github.com/conda-forge/conda-forge.github.io/tree/master/src/vote-results
+         folder.  The declaration PR should be merged by the first core member who is available to verify that
+         the requirements for the timeout have been met, based on their own personal records.
 
 ----
 
--- a/src/orga/governance.rst
+++ b/src/orga/governance.rst
@@ -104,6 +104,44 @@ To call for a standard vote, here is a template PR comment:
     This vote will end on {date}.
 
 ----
+:Quorum: All percentages below express *both* required participation, as a
+         fraction of the active core team, as well as the fraction of that
+         fraction who vote affirmatively on the issue. For example, in a vote
+         requiring 50%, with 18 active core members, at least 9 must vote, and
+         among them, there must 5 affirmative votes.
+
+         Because everyone is busy and it is difficult sometimes to get quorum,
+         there is a timeout, where the current participation level is taken for
+         what it is, and the percentage of affirmative votes is calculated from
+         whatever the vote total is at that time.  In order for a timeout to occur,
+         the vote must have:
+
+             * been open for at least 2 weeks
+             * been advertised on at least 3 separate occasions on the gitter core
+               channel (beginning of voting period, middle, and one day prior to
+               proposed timeout)
+             * been sent to core members via email.  Email reminders must have been
+               sent similarly to the gitter core channel: at least 3 times,
+               occurring as beginning of voting period, middle, and one day
+               prior to proposed timeout.
+
+         Extending the above example, if 9 people are required for a quorum, but
+         only 7 have voted, those 7 votes can form the basis of a completed vote
+         after the above conditions are met. 4 votes within those 7 would be
+         needed to pass the vote.
+
+         To post a timeout reminder, here is a template comment:
+
+          .. code-block:: md
+
+              @conda-forge/core
+              This vote falls under the {policy} policy, please vote and/or comment on this PR.
+              This vote needs {policy_percent} of core to vote yea to pass.
+              This vote presently has {current voters}, and needs {policy_percent * core} more for quorum.
+              It is proposed that this vote will timeout and be evaluated with the current votes in {days}, on {date}.
+              To vote please leave Approve (yea) or Request Changes (nay) reviews.
+
+----
 
 :CFEP Approval: When ready, the proposer may call for a vote on an
     existing conda-forge enhancement proposal (CFEP). This requires a
--- a/src/orga/governance.rst
+++ b/src/orga/governance.rst
@@ -85,8 +85,7 @@ Standard items are ones where public record and discourse is
 preferable. Sensitive voting items are ones where the results of the
 vote should remain private to the voters after the vote has occurred.
 Sensitive votes should take place on a secure anonymous voting platform
-(such as `Polys <polys.me>`_ or
-`the Helios voting system <https://vote.heliosvoting.org/>`_) in order to retain
+(we recommend `Polys <polys.me>`_) in order to retain
 anonymity. The email capability of your chosen voting platform should be used,
 and you should use the email list from
 https://github.com/conda-forge/conda-forge.github.io/blob/master/src/core.csv as
--- a/src/orga/governance.rst
+++ b/src/orga/governance.rst
@@ -110,7 +110,7 @@ To call for a standard vote, here is a template PR comment:
     This vote will end on {date}.
     
 ----
-:Posting results: To maintain the historical record, the outcome of any vote should be 
+:Posting results: To maintain the historical record, the outcome of any standard vote should be 
          recorded in the "vote-results" folder at 
          https://github.com/conda-forge/conda-forge.github.io/tree/master/src/vote-results
          
--- a/src/orga/governance.rst
+++ b/src/orga/governance.rst
@@ -86,8 +86,11 @@ preferable. Sensitive voting items are ones where the results of the
 vote should remain private to the voters after the vote has occurred.
 Sensitive votes should take place on a secure anonymous voting platform
 (such as `Polys <polys.me>`_ or
-`the Helios voting system <https://vote.heliosvoting.org/>`_) in order to
-retain anonymity.
+`the Helios voting system <https://vote.heliosvoting.org/>`_) in order to retain
+anonymity. The email capability of your chosen voting platform should be used,
+and you should use the email list from
+https://github.com/conda-forge/conda-forge.github.io/blob/master/src/core.csv as
+the authoritative list of emails to use.
 
 The default voting period is 1 week (7 days). This may be modified at
 the time when a vote is called, but may never be less than 24 hrs.
--- a/src/orga/governance.rst
+++ b/src/orga/governance.rst
@@ -139,6 +139,7 @@ To call for a standard vote, here is a template PR comment:
          time.  In order for a timeout to occur, the vote must have:
 
              * been open for at least 2 weeks
+             * been presented and discussed at a Core team meeting
              * been advertised on at least 3 separate occasions on the gitter core
                channel (beginning of voting period, middle, and one day prior to
                proposed timeout)
--- a/src/orga/governance.rst
+++ b/src/orga/governance.rst
@@ -135,7 +135,7 @@ To call for a standard vote, here is a template PR comment:
          must be affirmative.
 
          Because everyone is busy and it is difficult sometimes to get quorum,
-         votes not achieving quorum will eventually time out. When this happens,
+         votes not achieving quorum will eventually time out on their set end date. When this happens,
          the current participation level is taken for what it is, and the percentage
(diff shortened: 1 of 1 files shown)


src/orga/governance.rst
```diff
@@ -86,8 +86,11 @@ preferable. Sensitive voting items are ones where the results of the
 vote should remain private to the voters after the vote has occurred.
 Sensitive votes should take place on a secure anonymous voting platform
 (such as `Polys <polys.me>`_ or
-`the Helios voting system <https://vote.heliosvoting.org/>`_) in order to
-retain anonymity.
+`the Helios voting system <https://vote.heliosvoting.org/>`_) in order to retain
```
ocefpaf: Let's remove Helios. There should be only one voting platform to avoid confusion.

msarahan: Happy to.  That does expand the scope of this change, though.  If that's a sticking point for enough people, I'm also happy to leave it as-is.

CJ-Wright:
It may be good to set a minimum quorum. I'd rather not have a vote where one yea and the rest abstaining go through. These are decisions we need to stand behind as an organization, we need some form of consensus beyond one yes and a bunch of meh.

src/orga/governance.rst
ocefpaf: While I agree that it is kind of out of the scope part of the confusion started b/c, in helios, we only get a link to vote via e-mail. So maybe let's include that here and remove it. (Not even sure why people decided to keep it.)

mbargull:
I haven't yet read the PR.
This is copied over from the core chat just for myself as a reminder
<details>

> \[...\] it would make sense to have a minimum required number of voters (absolute or % of core) and relate the result on the number of voters, i.e., instead of
> 
> - X% Majority to pass
> have
> - N (or Y% of core) votes
> - and X% Majority of votes to pass
> 
> We should also offer an explicit "abstain" option to signal a non-yes/non-no vote that counts toward the total number of voters.
</details>

msarahan:
if people are really that "meh" about it, does it matter anyway?  I agree in principle.  I don't know how to express that practically.  Some things that really, really matter just don't inspire as much participation as they should.  For example, when we voted on changing the compilers way back when, the turnout was abysmal.

src/orga/governance.rst
scopatz: I think we had previously made changes so that we'd be platform agnostic to voting and not force us into a single platform.  CC @mariusvniekerk

msarahan:
To @CJ-Wright and others requesting a minimum quorum, please suggest guidelines for such a quorum.  That is the crux of this issue.  Our current minimum quorum corresponds to the vote fraction, which is applied twice, as described in this PR's text.  What would you change it to?  What is enough?  Is it a fraction?  An absolute number?

Votes happen because someone cares about something enough to put effort in.  If other people don't care enough to vote on it, should that really invalidate that person's initiative?  People don't vote for all kinds of reasons:

* don't feel technically qualified
* don't know a person who is being elected
* tied up with work / on hiatus
* can't be bothered
* didn't see the email / chat
* ???

I don't want people pushing crap through any more than anyone else, but we have a serious problem here.  We need mechanisms to reject bad stuff without letting other things die on the vine.  Maybe this is where the abstain vote comes in.  Maybe it's more of a comment that says "I don't have enough information to vote on this yet.  Could you please explain XYZ, or could Person XYZ please give a short introduction at the next meeting?"  We can't let things just not happen because of inaction.  That inaction is passive action, and it's miserable and disrespectful to the people who are expending effort and thought to get things done.  Pawning this off on "that's what emeritus is for" is irresponsible and unrealistic.  If people can't be bothered to vote, they sure as hell can't be bothered to actively check themselves in and out of emeritus status.

Rejection by inaction should not be a thing.

dougalsutherland:
I made some very small, hopefully uncontroversial, tweaks. Take a look and make sure you agree.

msarahan:
Thanks @dougalsutherland, those are good clarifications.

src/orga/governance.rst
msarahan: I'm going to revert that change.  I don't want to make this already-charged topic any more fraught.

msarahan:
I think the key thing here is that we are enforcing two things:

1. no one can try to sneak something past.  There must be loud communication on multiple fronts over a long time period, or else the vote cannot "time out."  I think this is key to the idea of the threat of anything passing with just 1 vote.  If anything passes with only one vote, we will all have been wildly irresponsible.
2. If nobody can be bothered to vote, it shouldn't block progress.

A huge part of this is changing the group's culture.  Right now, I think people stay silent when they are uncomfortable voting.  Really, they should say "I don't feel comfortable voting yet because X, Y, Z" and the person holding the vote needs to address those concerns.  Those concerns should be accepted in good faith, and best effort should be made to address them. That way, the vote is finite and does not end for ambiguous reasons.  People can still actively abstain if they really don't want to have an opinion, but  passively abstaining by not casting a vote should not deep-six the vote.

For example, for Forrest's election, I totally understand the concern that he's not well-known (yet).  It would have been wholly appropriate for anyone to say that: "I can't vote on Forrest yet because I don't know him well enough" - and it would then be on me as the vote organizer to figure out ways to help Forrest be known better, and the voting period would need to be extended accordingly.  I think people don't need to vote no unless they really disagree with something for some reason.  Here, that reason might be "I don't want anyone else from Anaconda on core."  OK, that's something I as vote organizer can't change.  Go ahead, vote no, but I hope you also communicate your concern because that's a larger governance issue that really should be aired and addressed.  If Forrest's election were to fail for this reason, then it would be appropriate to handle the larger governance issue (perhaps by defining explicit corporate membership), and reconsider Forrest afterwards depending on that outcome.

msarahan:
Yes, the burden of proof of notification is on the person who wants to declare the timeout, not on the community.

msarahan:
I added a documentation component for accountability.  I think adding a record of all of our votes (not just timed out ones) is good anyway.

scopatz:
Ping @conda-forge/core

src/orga/governance.rst
```diff
@@ -139,6 +139,7 @@ To call for a standard vote, here is a template PR comment:
          time.  In order for a timeout to occur, the vote must have:
 
              * been open for at least 2 weeks
+             * been presented and discussed at a Core team meeting
```
ocefpaf: We had that long before we created the Governance docs. It would be nice to have it again.
Not only we can call on people to vote but we can also answer clarification questions face-to-face. :+1:

src/orga/governance.rst
```diff
@@ -102,6 +106,68 @@ To call for a standard vote, here is a template PR comment:
...
+          .. code-block:: md
+
+              @conda-forge/core
+              This vote falls under the {policy} policy, please vote and/or comment on this PR.
+              This vote needs {policy_percent} of core to vote yea to pass.
+              This vote presently has {current_voters}, and needs {policy_percent * core - current_voters} more for quorum.
+              It is proposed that this vote will time out and be evaluated with the current votes in {days}, on {date}.
+              To vote please leave Approve (yea) or Request Changes (nay) reviews.
+              
+         To declare a vote "timed out," the person making such a declaration must post a pull-request adding
```
mbargull:
```suggestion
         To declare a standard vote "timed out," the person making such a declaration must post a pull-request adding
```
or
```suggestion
         To declare a public vote "timed out," the person making such a declaration must post a pull-request adding
```

src/orga/governance.rst
```diff
@@ -85,9 +85,13 @@ Standard items are ones where public record and discourse is
...
+in order to retain election integrity and anonymity. (We have used `Polys <polys.me>`_ 
+and `the Helios voting system <https://vote.heliosvoting.org/>`_, but are 
+open to any secure, anonymous system.) The email capability of your chosen 
+voting platform should be used for sending voting invitations and reminders,
+and you should use the email list from
+https://github.com/conda-forge/conda-forge.github.io/blob/master/src/core.csv as
+the authoritative list of emails to use.
 
 The default voting period is 1 week (7 days). This may be modified at
 the time when a vote is called, but may never be less than 24 hrs.
```
mbargull:
```suggestion
the time when a vote is called, but may never be less than 24 hrs.

Additional requirements may apply in case low turnouts have to be handled, see "Quorum" below.
```
^-- feel free to change the phrasing as you see fit.

This is just to remind people to plan ahead if they want timeouts to happen -> https://github.com/conda-forge/conda-forge.github.io/pull/970/commits/37f109730bb521186190c328045af1ea8a03de4b#diff-aeeef496313214bb0e1ff8c044612ad3R139 .

src/orga/governance.rst
mbargull: Not sure how you'd want "sensitive" votes to be handled.
Could just require Gitter + e-mail notification.
IDK if posting a PR for `vote-results/` with a redacted title/description makes much sense.

src/orga/governance.rst
```diff
@@ -102,6 +106,68 @@ To call for a standard vote, here is a template PR comment:
     To vote please leave Approve (yea) or Request Changes (nay) reviews.
     If you would like changes to the current language please leave a comment or push to this branch.
     This vote will end on {date}.
+    
+----
+:Posting results: To maintain the historical record, the outcome of any vote should be
```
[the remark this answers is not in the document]
msarahan: I read the difference between standard and sensitive as just whether the voting is anonymous or not.  I think vote totals in the historical record are all that is necessary - not names.  I see standard and sensitive votes as equivalent for this purpose.  Am I missing something here?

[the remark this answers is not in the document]
msarahan: In other words, *all* votes are public, it's just that some of them are voted on anonymously, while others take place in a way where names are shown.

msarahan:
I had one other idea here: instead of a sharp cutoff at a particular time, what about having a certain fraction of required votes decrease after a given time interval, like a roll-off?

For example, start at 3/4 of all core.  After 1 week, the required quorum is reduced by 1/8 of the needed members.  That continues every week until the quorum is met and the vote is tallied.

Any thoughts?  Or should we keep the cutoff?

beckermr:
I think simpler is better here maybe?

beckermr:
I had one more thought myself. Would it be too aggressive to automatically move people to emeritus after they miss a large number of votes, say 5 or 10?  

With the abstain option they don’t have to say yes or no anymore. And iiuic they can move back to active at any time.

msarahan:
I definitely agree with that concept, but I'm not sure how to do the bookkeeping.

beckermr:
Right. It could be something like 5 votes in a row but the bookkeeping is a pain indeed. We vote in more than one repo and idk if all platforms for voting support an option to see who votes but not how they voted.

isuruf:
If I understand correctly, a standard vote is 1 week, but the timeout occurs only after 2 weeks. For polys you can't extend the time right? If so, we'd have to redo the vote if we want to extend the time.

mwcraig:
> Right. It could be something like 5 votes in a row but the bookkeeping is a pain indeed.

As current president of our faculty senate I can vouch that this is a huge PITA. I haven't been actually keeping track of it even though we have an attendance list for each meeting.

I'd recommend keeping this separate for now.

src/orga/governance.rst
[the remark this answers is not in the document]
dougalsutherland: Hmm, this is interesting actually. The reason we set up sensitive votes as I remember it was so that a new core member wouldn't come in knowing "oh @dougalsutherland voted against me, weird"; it's marginally better to mask the name and know "oh two people voted against me", but still weird?

dougalsutherland:
Re: bookkeeping for sensitive votes: I think @mbargull brought up a good point in https://github.com/conda-forge/conda-forge.github.io/pull/970#discussion_r380528848. Do we want to track and make public that "three people voted against allowing X onto core", even if there are no names associated?

msarahan:
IMHO, yes, the history there is important.  I'd really like reasons to be posted anonymously, too.  This is all in the spirit of improvement. As long as people can be objective and not take things personally, rejection is a powerful tool towards refinement.

As the case of Forrest highlighted, some people were concerned about having too many corporate people on core (or too many from anaconda).  If we only capture the "no" vote, then we can't distinguish that as a vote against Forrest personally from a vote against adding another person from Anaconda.

For Forrest specifically, maybe people would vote no because they don't know him yet.  If we know that's the reason, we can take steps to improve things.

Similarly, if the concern is about Anaconda people on core, then that deserves a separate discussion outside of Forrest's election.

If all we have is a "no" vote, then we can't get at these concerns, and we can't work to address them.

jakirkham:
Those to me sound like different concerns. Namely...

1. Do we make all vote tallies public?
2. Do we track the reasons why people voted no?

It seems like we can do 2 without doing 1, which is what people seem concerned about.

msarahan:
The vote tallies are an aggregate statistic, and as such are anonymous whether the vote itself was anonymous or not.  I'm not sure why publishing the tallies is contentious.  Maybe someone could enlighten me?

As I've described above, I hope that people don't vote "no" when they want a revision.  Votes take a lot of people's time, and we should avoid re-doing votes as much as possible.  Revisions should ideally be made before the vote closes.

jakirkham:
AIUI that was what people were discussing [here]( https://github.com/conda-forge/conda-forge.github.io/pull/970#discussion_r380528848 ). Perhaps you can ask them 🙂

In any event, it seems this started as just making majority mean majority of those who voted as opposed to all of core (seemed agreeable, but we wanted to clearly specify it). Now the discussion has grown to revamping our voting system (generating a fair bit of side discussion). Are we biting off more than we can chew? 😉

src/orga/governance.rst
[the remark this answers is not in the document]
mbargull: We can discuss whether/how much of the "sensitive" vote results can or should be made public. But we should do that in a separate discussion/PR.

My intention is just to make sure we don't touch too many different topics here.
The description of "sensitive votes" has
> Sensitive voting items are ones where the results of the
vote should remain private to the voters after the vote has occurred.

which does not clarify whether the individual or the overall vote results are meant to "remain private to the voters". If we (in this first step, i.e., this PR) restrict the result posting to non-sensitive votes for now, we don't contradict any interpretation of that sentence.

(The anonymity part
> Sensitive votes should take place on a secure anonymous voting platform
in order to retain election integrity and anonymity.

is a little bit clearer due to the "retain election integrity and anonymity" phrase.)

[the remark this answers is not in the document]
msarahan: Fair enough.  Happy to merge and continue that later.

mbargull:
@isuruf 
> If I understand correctly, a standard vote is 1 week, but the timeout occurs only after 2 weeks. For polys you can't extend the time right? If so, we'd have to redo the vote if we want to extend the time.

Yes, that's correct. So you'd have to create a 2 week vote (plus advertizing it to core etc.) **upfront** to make it be able to timeout. That may not be ideal, but IMO we should let this PR pass and refine in subsequent PRs. Those subsequent PRs can be much easier put through since we'd have the initial governance change in effect by then.

isuruf:
You can't also stop it in the middle right?

mbargull:
That was why I wanted https://github.com/conda-forge/conda-forge.github.io/pull/970#discussion_r380547258 to be added.

mbargull:
> You can't also stop it in the middle right?

Right. Only after two weeks have passed and after the third time it has been advertized (which is right before it ends..

mbargull:
>> You can't also stop it in the middle right?
> 
> Right.

Oh, no, wait. It says "one day prior to proposed timeout". So it means timing out in the middle would be possible if you have a 4 week vote.

isuruf:
> Right. Only after two weeks have passed and after the third time it has been advertized (which is right before it ends..

After one week, there's nothing to do because we can't even see how many yes or nos. Just the number of yes+no+abstain. So, the polls will always go to 2 weeks.

dougalsutherland:
Right...this sounds like a tooling issue with Polys for our purposes here, which I don't see a nice way around.

It was fiddly, but was Helios better on this?

mbargull:
> After one week, there's nothing to do because we can't even see how many yes or nos. Just the number of yes+no+abstain.

I don't see that as much of a problem. Normally you'd only want to have the total number and decide to call for a timeout if only few have voted; you wouldn't want to rely on the actual votes.
(But yes, edge cases like 10 "Abstain"s + 1 "Yes" would be possible and of course questionable. If we ever have such a situation, we might have other more general problems, though.)
(A general required minimum number of votes as noted by some people may help there, too. We'd either need concrete suggestions for that or ... puts this in a subsequent "refinement" PR?!?)

mbargull:
> So it means timing out in the middle would be possible if you have a 4 week vote.

@msarahan, the reason for this is just that "proposed timeout" date is not said to be equal to the vote's end date. Was that intentional or can we just add a few words that clarify this?

src/orga/governance.rst
```diff
@@ -102,6 +108,68 @@ To call for a standard vote, here is a template PR comment:
...
+----
+:Quorum: All percentages below express *both* required participation, as a
+         fraction of the active core team, as well as the fraction of that
+         fraction who vote affirmatively on the issue. For example, in a vote
+         requiring 50%, with 18 active core members, at least 9 must vote;
+         if 9 vote, there must 5 affirmative votes. If 13 members vote, 7
+         must be affirmative.
+
+         Because everyone is busy and it is difficult sometimes to get quorum,
+         votes not achieving quorum will eventually time out. When this happens,
```
mbargull:
```suggestion
         votes not achieving quorum will eventually time out on their set end date. When this happens,
```
Would this suffice to address https://github.com/conda-forge/conda-forge.github.io/pull/970#issuecomment-588362001 ?

msarahan: I think this is a good clarification.  It's less that timing out is a separate process, and more that it is an inevitable end, but in order to reach that end, the notification requirements must be met.

mbargull: Thanks!
@isuruf, does that address some of your concerns?

isuruf: Not really. We should then mention that secret/sensitive votes should have a minimum of 2 weeks.

mbargull: Hmm, we can't do that because of, e.g., https://github.com/conda-forge/conda-forge.github.io/pull/970/files#diff-aeeef496313214bb0e1ff8c044612ad3L175 when we would (unfortunately) have to have short sensitive votes.

So, I guess you mean https://github.com/conda-forge/conda-forge.github.io/pull/970#discussion_r380547258 is not clear enough -- do you have any alternative phrasing suggestions?

isuruf: If somebody had a 1 week vote and didn't get a quorum, it'll be in limbo, because after 1 week you can't vote, so the conditions for timing out cannot be fulfilled.

mbargull: Correct. That's what my "Additional requirements may apply \[...\]" hint at: If you want to make use of the new timeout rule, you have to meet the additional requirements (2+ weeks etc.). Shorter votes will be handled as is currently the case (i.e., no timeout can be applied).

mbargull: @isuruf, are you fine with the current state of the proposal then?
We are currently sitting at `14` out of `15` required approvals.

src/orga/governance.rst
```diff
@@ -85,13 +85,19 @@ Standard items are ones where public record and discourse is
...
+open to any secure, anonymous system.) The email capability of your chosen 
+voting platform should be used for sending voting invitations and reminders,
+and you should use the email list from
+https://github.com/conda-forge/conda-forge.github.io/blob/master/src/core.csv as
+the authoritative list of emails to use.
 
 The default voting period is 1 week (7 days). This may be modified at
 the time when a vote is called, but may never be less than 24 hrs.
 
+Additional requirements may apply in case low turnouts have to be handled, see "Quorum" below.
```
isuruf:
```suggestion
Additional requirements may apply in case low turnouts have to be handled, see "Quorum" below.
For sensitive votes, this must be planned ahead when deciding the duration of the vote.
```

mbargull: Why the restriction on "sensitive" votes?

isuruf: Other votes can be extended. Sensitive votes have to be redone.

mbargull: Hm.. Not really, actually. The document only talks about vote end dates but does not offer the possibility the change/extend those.
De facto we just have *blatantly ignored the end dates* on many occasions -- something I don't like and would like us to change in the future. It's just not good practice and does not seem professional.
If we want to make use of an extended voting period, we should just open the vote for a month instead of two weeks then, IMHO. (Otherwise we should amend the governance document to clarify what "end date" is supposed to mean for standard votes then.)

mbargull:
This vote has passed on `2020-02-20T18:33+00:00` with `15/19 = 78.9%` of `conda-forge/core` approving.
(^-- Given the new governance changes from this PR, this should be recorded in a new file under `https://github.com/conda-forge/conda-forge.github.io/tree/master/src/vote-results/`.)

If you have any further change requests/refinements to this, please feel encouraged to issue follow-up PRs.

Thanks everyone!

msarahan:
Thanks.  I will put up a separate PR for the folder and this result.

msarahan:
PR at https://github.com/conda-forge/conda-forge.github.io/pull/987

PR #1031, opened 2020-04-17
Redirect ReadTheDocs 404's to current docs

jakirkham:
Follows [this documentation]( https://docs.readthedocs.io/en/stable/guides/custom-404-page.html#use-a-custom-404-not-found-page-on-my-project ) to setup a 404 page. The idea being it will redirect any other pages we don't provide (that would otherwise 404) to our current doc page. Reuses the same `index.html` to create the `404.html`.

Publishing to here: https://conda-forge.readthedocs.io/en/jakirkham-rtd_redirect_404

The idea is this ~~*should* (though currently doesn't)~~ does redirect: https://conda-forge.readthedocs.io/en/jakirkham-rtd_redirect_404/foobar.html

### DIFF
--- a/rtd/conf.py
+++ b/rtd/conf.py
@@ -12,6 +12,7 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
+import shutil
 import sys
 import os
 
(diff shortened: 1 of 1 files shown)


jakirkham:
cc @beckermr

jakirkham:
Ok got this working now. This is ready for review! 😄

beckermr:
This looks great! I think with one more hack taken from the dask-ml stuff we can make it not load the full old RTD template which should make it snappier. I'll push that in another PR.

jakirkham:
Thanks! Sounds good 😄

PR #1032, opened 2020-04-17
Ignore `*.profraw` files built by docs

jakirkham:


jjhelmus:
Are these coming from a macOS build or Linux?  The former has been reported at conda-forge/python-feedstock#339

jakirkham:
They were showing up on macOS. Though had assumed these were a Sphinx thing not a Python thing.

PR #1033, opened 2020-04-17
ENH snappier rtd

beckermr:
This change makes it a bit snappier since it is now plain text.

cc @jakirkham


jakirkham:
Awesome! Thanks 😄

PR #1037, opened 2020-04-19
sphinx-3 breaks recommonmark md file parsing

croth1:
It seems like sphinx-3 and recommonmark do not play well together for us. The minutes page is not getting populated anymore. see: https://conda-forge.org/docs/minutes/00_intro.html


scopatz:
Thanks!

PR #1041, opened 2020-04-25
ENH automagic version

beckermr:


src/conf.py
```diff
@@ -57,17 +58,17 @@
 
 # General information about the project.
 project = u'conda-forge'
-copyright = u'2016-2019, conda-forge'
```
croth1, line 60: Does it make sense to also generate the current year automatically?

beckermr: good idea

beckermr: done.

PR #1048, opened 2020-04-29
Add minutes from last quarter

ericdill:


croth1:
In a recent PR I moved the minutes from src/minutes to src/orga/minutes. I hope you didn't mind that.

PR #1058, opened 2020-05-09
ENH announce CFEP-13

beckermr:
This PR has the announcement for the start of the CFEP-13 rollout.

### DIFF
--- a/src/user/announcements.rst
+++ b/src/user/announcements.rst
@@ -4,6 +4,33 @@ Announcements
 2020
 ----
 
+:2020-05-09: New Staging Process for ``anaconda.org`` Uploads
+
+    Starting this week, we are changing the way we upload packages to ``anaconda.org``. 
+    We will move from direct uploads to the ``conda-forge`` ``main`` channel to using a 
+    staging organization/channel combined with a copy request from the staging channel to 
+    the production channel. This new process will allow us to perform some validation on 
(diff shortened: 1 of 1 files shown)


beckermr:
@conda-forge/core any comments here would be greatly appreciated!

croth1:
I agree with @xhochy that using both the blog and the *Announcements* section will make it hard for conda-forge users to get the message.
In fact I am not sure how many people actually know that we have this, let alone follow it.

The blog has the advantage that it offers an RSS feed, which would be my preferred way to get news from conda-forge, but I am not sure how well that generalizes.

I would suggest following strategy:
- Finishing up #1049 - otherwise people may have a hard time finding the blog.
- Migrating the announcement section fully to the new blog, I think we should keep the history.
- link to the blog in the former announcement section.
- Make a poll on our twitter account with following options: I am interested in subscribing to the latest news from conda-forge and I would prefer:
  - RSS feed
  - Twitter+RSS feed
  - Mailing list
  - Other, please comment

In the end if it's not useful to the community, the time is not time well spent, IMO.

beckermr:
@xhochy @croth1 

I think the blog and the announcements serve two different purposes. 

The announcements section is linked as `NEWS` on the main page and is effective the changelog for conda forge. It contains technical changes and some documentation about them for users and maintainers.

The blog is for us to present long-form content that go beyond technical documentation.

PR #1075, opened 2020-05-28
Update "Current Members of Core" section

marcelotrevisani:

### DIFF
--- a/src/core.csv
+++ b/src/core.csv
@@ -1,3 +1,4 @@
+github_username,email,name
 beckermr,becker.mr@gmail.com,Matthew R. Becker
 cj-wright,cjwright4242@gmail.com,Christopher J. 'CJ' Wright
 djsutherland,djs@djsutherland.ml,D.J. Sutherland
--- a/src/orga/governance.rst
+++ b/src/orga/governance.rst
@@ -310,7 +310,6 @@ Current Members of Core
 In alphabetical order,
 
 * Anthony Scopatz, @scopatz
-* Björn Grüning, @bgruening
 * Christopher J. "CJ" Wright, @CJ-Wright
 * Dougal J. Sutherland, @dougalsutherland
 * Eric Dill, @ericdill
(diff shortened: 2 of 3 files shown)


jjhelmus:
A few of the member on this list are emeritus.  Should we split this list to show those member under a separate heading?  
The [core.csv](https://github.com/conda-forge/conda-forge.github.io/blob/master/src/core.csv) and [emeritus.csv](https://github.com/conda-forge/conda-forge.github.io/blob/master/src/emeritus.csv files in this repository give list of the current core members and emeritus members.

CJ-Wright:
You might consider using this as well (it has current and emeritus) https://github.com/conda-forge/by-the-numbers/blob/master/plots/current-core-devs.png although it needs to be updated

marcelotrevisani:
@jjhelmus @CJ-Wright 
should we split the list as @jjhelmus mentioned ? 

> You might consider using this as well (it has current and emeritus) https://github.com/conda-forge/by-the-numbers/blob/master/plots/current-core-devs.png although it needs to be updated

I will do that as well, thanks

marcelotrevisani:
I believe all the core and emeritus members are in this PR
Did I miss someone?

CJ-Wright:
At some point we might consider just jinja templating this.

src/orga/governance.rst
```diff
@@ -309,27 +309,37 @@ Current Members of Core
...
 * Mike Sarahan, @msarahan
-* Anthony Scopatz, @scopatz
-* Patrick Snape, @patricksnape
+* Min Ragan-Kelley, @minrk
 * Patrick Sodré, @sodre
-* Dougal J. Sutherland, @dougalsutherland
 * Peter K. G. Williams, @pkgw
-* Christopher J. "CJ" Wright, @CJ-Wright
+* Uwe L. Korn, @xhochy
+* Wolf Vollprecht, @wolfv
```
scopatz, line 332: Isn't alphabetical order usually by last name?

marcelotrevisani: I think if the name was written as ``Surname, the first name`` I would agree with you
but I think as we wrote the name in full that might be a bit odd (at least for me) to sort it looking for the last name and writing the first name at the beginning

src/core.csv
```diff
@@ -1,3 +1,4 @@
+github_username,email,name
```
CJ-Wright, line 1: I'm not certain if this will play well with the voting system.

marcelotrevisani: How can we check it?

CJ-Wright: You could try to make a dummy vote in polys

marcelotrevisani: I tested and it worked fine

marcelotrevisani: I tested with the csv header and add you and me to a pool
it seems to be working fine

PR #1077, opened 2020-06-07
Update adding_pkgs.rst

AndrewAnnex:
removes typo from "Copying test files" section in docs. tests and example .py file can't be also in the requires section

### DIFF
--- a/src/maintainer/adding_pkgs.rst
+++ b/src/maintainer/adding_pkgs.rst
@@ -556,7 +556,7 @@ You can include files required for testing with the ``source_files`` section:
       imports:
         - package_name
       requires:
-        - pytest tests test_pkg_integration.py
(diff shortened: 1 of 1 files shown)


src/maintainer/adding_pkgs.rst
```diff
@@ -556,7 +556,7 @@ You can include files required for testing with the ``source_files`` section:
       imports:
         - package_name
       requires:
-        - pytest tests test_pkg_integration.py
```
beckermr, line 559: We should add this line back as a commands section.

AndrewAnnex: done, placed after source_files

jakirkham:
Thanks Andrew! 😀

Issue #1078, opened 2020-06-08
pydstool-feedstock, offer of new maintainer


repagh:
Issue: Current maintainer of pydstool has moved on, see quote from e-mail:

> Hey,
>  I no longer maintain that feedstock since there was little to no gain and the Julia package ecosystem is already much more advanced in this area that we've dropped our usage of pydstool. Best of luck to you.

I hope this is the right forum for this issue. 

I am OK with taking over the feedstock, can you add me as maintainer?

repagh / René van Paassen

croth1:
You can add yourself to the maintainer list in the [meta.yaml](https://github.com/conda-forge/pydstool-feedstock/blob/master/recipe/meta.yaml#L61) and submit this as a PR. The current maintainer or core (e.g. by pinging) [at]conda-forge/core can merge the PR and our infrastructure will make sure that you get writing permission.

repagh:
Thanks, I see this is much clearer now #1079

PR #1079, opened 2020-06-08
add guidelines on how to take over maintenance

croth1:
adds guideline for taking over maintainer role. (see also https://github.com/conda-forge/conda-forge.github.io/issues/1078)

### DIFF
--- a/src/orga/guidelines.rst
+++ b/src/orga/guidelines.rst
@@ -121,27 +121,10 @@ by default. However, the package remains available so as to avoid the problems l
 To get your package relabeled as ``broken`` instead of ``main``, please refer to :ref:`maint_fix_broken_packages`.
 
 
-.. _becoming_maintainer:
-
 Becoming a maintainer
 =====================
 
 Conda-forge is a community project and it can therefore happen that feedstocks become temporarily abandoned.
-You can join the maintainer team by following steps:
+You can join the maintainer team of a feedstock by adding your github-id to the ``recipe-maintainers`` section in the recipe's ``meta.yaml``. 
+Please refer to :ref:`maint_updating_maintainers` for detailed instructions.
 
-1) Add your github-id to the ``recipe-maintainers`` section at the bottom of the ``recipe/meta.yaml`` file in the feedstock:
-
-  .. code-block:: yaml
-
-    extra:
-      recipe-maintainers:
-        - current-maintainer
-        - your-github-id
-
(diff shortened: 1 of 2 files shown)


src/orga/guidelines.rst
```diff
@@ -119,3 +119,25 @@ drops the "main" label and replaces it with "broken". This still makes it unavai
...
+
+  .. code-block:: yaml
+
+    extra:
+      recipe-maintainers:
+        - current-maintainer
+        - your-github-id
+
+2) Commit and push the change to your fork and open a :term:`PR` against the feedstock you want to become a maintainer of.
+
```
jakirkham, line 140: Would add re-render the feedstock since this updates the codeowners and maintainer listing in the Readme.

croth1: Good catch! I added an extra step.

croth1:
By the way, so you think guidelines is the right place to put this? It would also fit it the maintainer section.

src/orga/guidelines.rst
```diff
@@ -119,3 +119,29 @@ drops the "main" label and replaces it with "broken". This still makes it unavai
...
+.. _becoming_maintainer:
+
+Becoming a maintainer
+=====================
+
+Conda-forge is a community project and it can therefore happen that feedstocks become temporarily abandoned.
+You can join the maintainer team by following steps:
+
+1) Add your github-id to the ``recipe-maintainers`` section at the bottom of the ``recipe/meta.yaml`` file in the feedstock:
```
scopatz, line 132: Should be `1.` etc for rST

jakirkham:
> By the way, so you think guidelines is the right place to put this? It would also fit it the maintainer section.

Either sound ok to me. Though I think you are the expert on doc organization at this point 😄

src/orga/guidelines.rst
jakirkham: Thanks! 😄

CJ-Wright:
Maybe put in a cross reference?

croth1:
> Maybe put in a cross reference?

I like that idea! I pushed the step-by-step instructions to the maintainer section and referenced it in the guidelines. That's consistent with how we did things before :-)

src/orga/guidelines.rst
croth1: Fixed! Thanks!

scopatz:
Thanks!

PR #1105, opened 2020-07-15
announce the move to clang 10

ocefpaf:
Ping @isuruf

### DIFF
--- a/src/user/announcements.rst
+++ b/src/user/announcements.rst
@@ -7,7 +7,7 @@ Announcements
 
 :2020-07-16: Moving from clang 9 to clang 10
 
-    conda-forge is moving to clang 10!
+    conda-forge is moving to clang 10 on macOS!
     Check the release `notes <https://releases.llvm.org/10.0.0/tools/clang/docs/ReleaseNotes.html#what-s-new-in-clang-10-0-0>`_
     for what is new, breaking, or deprecated.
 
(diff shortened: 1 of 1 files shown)


src/user/announcements.rst
```diff
@@ -5,6 +5,12 @@ Announcements
 ----
 
 
+:2020-07-15: Moving from clang 9 to clang 10
```
ocefpaf, line 8: I'm not sure how ReST works, should we have only a single `2020-07-15` header? If so just offsetting by the same amount of spaces the CFEP-18 would work?

src/user/announcements.rst
```diff
@@ -5,6 +5,12 @@ Announcements
 ----
 
 
+:2020-07-16: Moving from clang 9 to clang 10
+
+    conda-forge is moving to clang 10!
```
xhochy, line 10: This is probably only for macOS, we don't use it somewhere else.
```suggestion
    conda-forge is moving to clang 10 on macOS!
```

src/user/announcements.rst
ocefpaf: I like things that solves themselves by just waiting a little bit :smile:

ocefpaf:
@isuruf is this OK to be merge? I'll send the other one for openblas as soon as we get this one in.

ocefpaf:
Alright, the updates were approved. Merging this.

PR #1122, opened 2020-07-31
JupyterLab Extension Docs

scopatz:
This came out of conversations that I have been having with @saulshanabrook & @ilanschnell about what the proper way to package JupyterLab extensions with conda is.  Right now there is no consistency in conda-forge in terms of how extensions are packaged. This is the first step in addressing that, where we are trying to write down what the right way of doing it actually is.

CC @jakirkham - you might be interested in this.

### DIFF
--- a/src/maintainer/knowledge_base.rst
+++ b/src/maintainer/knowledge_base.rst
@@ -323,6 +323,52 @@ At the time of writing, above is equivalent to the following,
     there are ``numpy`` packages for ``1.7`` available for Python 2.7 in the channel.
 
 
+.. _jupyterlab_extension:
+
+JupyterLab Extensions
+---------------------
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Thanks for tackling this!

Some remarks on lab extensions

 - many are not python packages at all and are only available on NPM.
 - some that *are* python packages already include the tjz file as data file.
 - this will probably change with lab 3.0 which will have the dynamic extension system.

jakirkham:
Thanks Anthony! Sorry I missed this before.

Sylvain, those sound like reasonable improvements, do you think it would be worthwhile to move those to an issue or add a follow-up PR so we can better track them?

scopatz:
Yeah, I expect these docs to change for JLab 3, which may entail separate / updated instructions

PR #1130, opened 2020-08-07
DOC Describe azure store_build_artifacts flag for conda-forge.yml

ryanvolz:
Flag was added to conda-smithy with https://github.com/conda-forge/conda-smithy/pull/1355.


ryanvolz:
> You need to modify the files in src, not docs.

D'oh! And I thought I'd avoided making that mistake by not editing the html... one sec.

PR #1148, opened 2020-08-30
BUG write RST

beckermr:
cc @ericdill


ericdill:
maybe just combine this with the "ABOUT" section and call it a day?

PR #1179, opened 2020-11-02
BUG add env vars

beckermr:
The env vars need adjusting and do not appear to be set. This should fix that.

### DIFF
--- a/.ci_scripts/update_docs
+++ b/.ci_scripts/update_docs
@@ -35,7 +35,7 @@ if [ -n "$GH_TOKEN" ]; then
 
     # Only commit the docs for commits on master.
     # GITHUB_HEAD_REF is empty on forks
-    if [[ "$GH_REF" == "ref/heads/master" ]] && [[ "$GH_HEAD_REF" == "" ]]; then
+    if [[ "$GH_REF" == "ref/heads/master" ]] && [[ "$GH_ACTOR" == "conda-forge" ]]; then
         git add -A :/
         git commit -m "Re-ran make.py. [ci skip]" || true
         git remote add pushable https://${GH_TOKEN}@github.com/conda-forge/conda-forge.github.io.git/ > /dev/null
--- a/.ci_scripts/update_docs
+++ b/.ci_scripts/update_docs
@@ -31,11 +31,11 @@ if [ -n "$GH_TOKEN" ]; then
     cd ..
     mv docs-README docs/README
 
-    git diff
+    # git diff
 
     # Only commit the docs for commits on master.
     # GITHUB_HEAD_REF is empty on forks
-    if [[ "$GITHUB_REF" == "ref/heads/master" ]] && [[ "$GITHUB_HEAD_REF" == "" ]]; then
+    if [[ "$GH_REF" == "ref/heads/master" ]] && [[ "$GH_HEAD_REF" == "" ]]; then
         git add -A :/
         git commit -m "Re-ran make.py. [ci skip]" || true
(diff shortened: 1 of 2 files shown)


.ci_scripts/update_docs
```diff
@@ -31,11 +31,11 @@ if [ -n "$GH_TOKEN" ]; then
     cd ..
     mv docs-README docs/README
 
-    git diff
+    # git diff
 
     # Only commit the docs for commits on master.
     # GITHUB_HEAD_REF is empty on forks
-    if [[ "$GITHUB_REF" == "ref/heads/master" ]] && [[ "$GITHUB_HEAD_REF" == "" ]]; then
+    if [[ "$GH_REF" == "ref/heads/master" ]] && [[ "$GH_HEAD_REF" == "" ]]; then
```
chrisburr, line 38: I think these need to be set in the workflow YAML using the `${{ .... }}` syntax?

chrisburr: Ah I'm wrong, `GITHUB_REF` should be set: https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables

beckermr: Maybe. I am having trouble with it so I put them in the yaml.

.github/workflows/deploy.yml
```diff
@@ -35,3 +35,5 @@ jobs:
           source ./.ci_scripts/update_docs
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REF: ${{ github.ref }}
+          GH_ACTOR: ${{ github.actor }}
```
chrisburr, line 39: Isn't this the user that triggered the CI (likely the person who merged). I think this gives you `conda-forge/conda-forge.github.io`:
```suggestion
          GH_ACTOR: ${{ github.repository }}
```

beckermr: IDK. I was looking at this: https://github.community/t/dont-run-actions-on-pull-request-from-fork/17560

beckermr: I changed this around a bit.

.github/workflows/deploy.yml
```diff
@@ -35,3 +36,5 @@ jobs:
...
```
beckermr, line 40:
```suggestion
```

beckermr:
OK @chrisburr. Take a look now.

.ci_scripts/update_docs
```diff
@@ -1,52 +1,43 @@
...
+mv docs-README docs/README
+
+# git diff
+
+# Only commit the docs for commits on master.
+# GITHUB_HEAD_REF is empty on forks
+if [[ "$GH_REF" == "ref/heads/master" ]] && [[ "$GH_TOKEN" != "" ]]; then
+    git add -A :/
+    git commit -m "Re-ran make.py. [ci skip]" || true
+    git remote add pushable https://${GH_TOKEN}@github.com/conda-forge/conda-forge.github.io.git/ > /dev/null
```
chrisburr, line 34: I think this will push from core member's forks if they push commits to master (e.g. `chrisburr/conda-forge.github.io.git`)

beckermr: Those forks won't have a valid token IIUIC. We are using the builtin token from github which is scoped to a single repo.

beckermr: It's not working anyways. Here comes another PR!

PR #1184, opened 2020-11-12
Update knowledge_base.rst.txt

megies:
just a tiny typo fix


megies:
My bad, wrong file edited

PR #1196, opened 2020-11-23
Add MacStadium to inst_partners.yaml

chrisburr:
@SylvainCorlay Is this data used anywhere yet?

### DIFF
--- a/src/inst_partners.yaml
+++ b/src/inst_partners.yaml
@@ -93,9 +93,17 @@ Works on Arm:
     - infrastructure
 
 OVHCloud:
-   Description: OVH Provides a Windows Build VM
+   Description: OVH provides a Windows build VM
    Site URL: https://www.ovh.com/
    Logo URL:
      - https://www.frenchtechbordeaux.com/wp-content/uploads/2019/08/LogoOVHCloud_Vertical-Blue2.png
+   contribution type:
+     - infrastructure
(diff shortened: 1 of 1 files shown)


src/inst_partners.yaml
```diff
@@ -93,9 +93,17 @@ Works on Arm:
...
    Logo URL:
      - https://www.frenchtechbordeaux.com/wp-content/uploads/2019/08/LogoOVHCloud_Vertical-Blue2.png
    contribution type:
      - infrastructure
+
+MacStadium:
+   Description: MacStadium provides dedicated Mac mini hardware
+   Site URL: https://www.macstadium.com/
+   Logo URL:
+     - https://uploads-ssl.webflow.com/5ac3c046c82724970fc60918/5c019d917bba312af7553b49_MacStadium-developerlogo.png
```
beckermr, line 107: We'll need some agreement to use their logo. Otherwise, we should comment out or leave this out.

chrisburr: They've explicitly requested that we use this URL somewhere.

beckermr: We do not have a page yet, so we'll need to make one.

SylvainCorlay: yes indeed this data is not used yet.

PR #1201, opened 2020-12-02
add url for lab49 logo

CJ-Wright:


CJ-Wright:
@beckermr are `.svg`s acceptable?

beckermr:
It gets put into an html image tag. Should be ok I think.

CJ-Wright:
Thank you!

PR #1203, opened 2020-12-03
DOC add announcement about artifact validation

beckermr:
This PR adds an announcement about artifact validation.

### DIFF
--- a/src/user/announcements.rst
+++ b/src/user/announcements.rst
@@ -4,14 +4,26 @@ Announcements
 2020
 ----
 
+:2020-12-02: Artifact Validation
+
+    In an effort to better secure ``conda-forge``, we are developing a process to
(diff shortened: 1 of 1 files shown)


jakirkham:
This looks good. Thanks Matt! 😄

Maybe you mentioned this this morning and I forgot, but is there any way for maintainers to see logs from package validation or get updates about it? Am wondering if there are ways we can help maintainers be more proactive here to minimize surprise and start fixing issues early. WDYT? 🙂

beckermr:
Yes. My plan is to start sending issues to feedstocks when things fail and bumping them on admin requests when things should be marked as broken. 

For now, people can checkout this list of invalid things: https://github.com/conda-forge/artifact-validation/blob/master/scan_data/invalid_packages.yaml

This is not a public API so I don't want to advertise too widely.

PR #1211, opened 2020-12-09
Remove --no-deps mention

xhochy:
This is not needed anymore as this is added via environment flags in the build process and tools like https://github.com/conda-incubator/grayskull don't add it to the script anymore. We should also remove it from the docs to avoid confusion on whether this is required.


beckermr (approved):

jakirkham:
Thanks Uwe! 😄

PR #1222, opened 2020-12-16
ENH better packages

beckermr:

### DIFF
--- a/404.html
+++ b/404.html
@@ -69,7 +69,7 @@
                         <a class="page-scroll" href="https://conda-forge.org/#contribute">Contribute</a>
                     </li>
                     <li>
-                        <a href="https://conda-forge.org/feedstocks">Packages</a>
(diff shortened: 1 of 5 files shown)


404.html
```diff
@@ -69,7 +69,7 @@
                         <a class="page-scroll" href="https://conda-forge.org/#contribute">Contribute</a>
                     </li>
                     <li>
-                        <a href="https://conda-forge.org/feedstocks">Packages</a>
+                        <a href="https://conda-forge.org/https://conda-forge.org/feedstock-outputs">Packages</a>
```
pkgw, line 72: Is this correct?

beckermr: yup.

beckermr: oh woops

beckermr: Good catch!

PR #1236, opened 2021-01-04
DOC address comments from @isuruf

beckermr:


beckermr:
A minor change from what you posted @isuruf to make sure it was clear our default is strict regardless of the conda default.

isuruf (approved):

PR #1244, opened 2021-01-13
Add an index of CFEPs

### LINKED ISSUE #1241: Add index of CFEPs
I think it would be nice to have an index of CFEPs in the main documentation.

Even when the main content of a CFEP has been included in the docs, CFEPs provide useful information. They are not very easy to find today. An index would ease that.
It would be good if this was automated but I don't really see how it would fit with the sphinx documentation. On the other side, there aren't that many CFEPs. Updating the index manually should be fine.

If there is interest, I could make a PR.

beenje:
Fix #1241 

Note that the index needs to be maintained and updated manually.

### DIFF
--- a/.ci_scripts/generate_cfep_index.py
+++ b/.ci_scripts/generate_cfep_index.py
@@ -1,0 +1,66 @@
+import re
+import requests
+from dataclasses import dataclass
+from pathlib import Path
+
+REPO_CONTENTS = "https://api.github.com/repos/conda-forge/cfep/contents/"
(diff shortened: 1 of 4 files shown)


beckermr (commented):
Can we generate this list automatically from the CFEP repo?

beenje:
Will try to propose a small script!

beenje:
I added a script that should append the list of CFEPs automatically to the `cfep-index.rst` file.

jakirkham:
Thank you for all of these doc improvements! 😄

PR #1245, opened 2021-01-14
initial pass at arch diagrams both abstract and implementation

CJ-Wright:
The idea is to capture all the moving pieces at both in an abstract,
service oriented way and from an implementation perspective.

CC: @beckermr @kruttik-lab49 @k504866430


kruttik-lab49:
It might be helpful to add numbered labels to arrows- e.g
```
staged_recipes >> Edge(label="1. build") >> cis 
```

This ofcourse has a downside in that it implies a certain directionality. 

Overall I think the numbered arrows might be useful.

arch_diagrams/create_diagrams.py
```diff
@@ -0,0 +1,81 @@
...
+        validation = Heroku(label='pkg validation')
+        other_cmds = Heroku(label='webservice commands')
+
+    rerender = GithubActions(label='feedstock rerender')
+    other_cmds >> rerender
+
+    create_feedstock = GithubActions(label='create feedstock repo')
+    autotick = GithubActions(label='autotick bot')
+
+    anaconda_server = Storage(label='Anaconda Servers')
```
chrisburr, line 63: I'm not sure if this is too much detail but I think this should contain more detail about how the channel is implemented. Perhaps three components:
* anaconda.org
* CDN mirroring with repodata patches being an input
* The publicly visible channel

arch_diagrams/create_diagrams.py
```diff
@@ -0,0 +1,81 @@
...
+
+    internet = Datacenter(label='new version numbers')
+    validation = Heroku(label='pkg validation')
+
+    admin_migrations = CircleCI('admin migrations')
+
+
+    staged_recipes >> cis >> staged_recipes
+    staged_recipes >> linter >> staged_recipes
+    staged_recipes >> create_feedstock >> new_feedstock >> cis >> validation >> anaconda_server
```
chrisburr, line 32: I'm not sure if package validation needed for the conceptual diagram as I see it as more of an implementation detail. Perhaps "Anaconda Servers" should be "conda channel"?

beckermr: We maintain the validation with infrastructure we control so it needs to be here.

arch_diagrams/create_diagrams.py
beckermr: This is all stuff anaconda maintains and so we shouldn't have it here. To us anaconda.org is a free service with a big latency for uploads.

arch_diagrams/create_diagrams.py
chrisburr: I agree it needs to be the "Conda-Forge Implementation" diagram but I'm not sure it needs to be in the "Conda-Forge Conceptual" diagram though I don't feel that strongly about it if others disagree.

arch_diagrams/create_diagrams.py
chrisburr: I think it's useful to be here so someone can look at this and understand what is might be wrong with conda-forge for a given symptom. Questions and issues with the CDN sync common enough that I think it's worthwhile acknowledging it exists in the conda-forge documentation. Again, I don't feel that strongly about it if others disagree.

arch_diagrams/create_diagrams.py
beckermr: ah sure agreed!

arch_diagrams/create_diagrams.py
beckermr: sure - we can put a big anaconda box around things

arch_diagrams/create_diagrams.py
CJ-Wright: Hmm, validation might be an independent concept, since it isn't formally owned by any of the other pieces. Also the fact that we validate our packages is important to the running of conda-forge, since is a barrier to some potential attacks. I would call the implementation detail that this is run off of the heroku server for instance. I'm happy to be turned around on this though.

PR #1246, opened 2021-01-20
Update build number in MPI example

chrisburr:
I don't think the 1000 build number is needed any more unless there is a quirk on MPI which I'm missing?


beckermr:
@minrk You built a lot of the MPI stuff here. Can we go back to zero-based numbers?

minrk (approved):

minrk:
I think the 1000 is no longer useful

beckermr:
Thanks @minrk !

PR #1249, opened 2021-01-22
knowledge_base: add missing cudnn key

duncanmmacleod:
Add missing `cudnn` key in YAML block describing `conda_build_config.yaml` for CentOS 7

cc @chrisburr @beckermr


beckermr:
Thank you!

PR #1250, opened 2021-01-22
Add FAQ to explain CF's CUDA compiler is not intended to be used by end users

leofang:
As mentioned in https://github.com/conda-forge/conda-forge.github.io/pull/1226#discussion_r554135009 and https://github.com/conda-forge/cudatoolkit-feedstock/issues/27#issuecomment-765497597 (just two most recent examples) we got asked numerous time why this is not possible. I hope this helps clarify a bit, although I am not sure how many users will actually check it out...

cc: @jaimergp @jakirkham @kkraus14


beckermr (commented):
From the FAQ, it sounds like if you do install the toolkit, this is possible. Do we want to provide instructions somewhere on how to do that?

leofang:
No, because it's well documented in CUDA Toolkit's documentation. Are you saying we should refer users to that? If so we can certainly add a link.

jakirkham:
Adding a link seems like a good idea 🙂

Thanks for working on the Leo 😄

leofang:
Added. PTAL.

src/user/faq.rst
```diff
@@ -71,4 +71,10 @@ FAQ
 
     x86_64-apple-darwin13.4.0-clang: No such file or directory
     
-  are a telltale sign that you are lacking compilers.
\ No newline at end of file
+  are a telltale sign that you are lacking compilers.
+
+:ref:`(Q) <faq_cuda_compiler_header>` **How can I compile CUDA (host or device) codes in my environment?**
+
+  Unfortunately, this is not possible with Conda-Forge's current infrastructure (``nvcc``, ``cudatoolkit``, etc) if there is no local CUDA Toolkit installation. In particular, the ``nvcc`` package provided on Conda-Forge is a *metapackage* wrapping an actual ``nvcc`` compiler for our CI to use; it does not contain the full ``nvcc`` compiler toolchain. One of the reasons is that CUDA headers like ``cuda.h``, ``cuda_runtime.h``, etc, are not redistributable according to NVIDIA's EULA, which are needed at compile time. Likewise, the ``cudatoolkit`` package only contains CUDA runtime libraries and not the compiler toolchain.
```
jaimergp, line 78: Nitpicking here, feel free to ignore:

In the maintainer bits I referred to `nvcc` as a *wrapper package* since it doesn't provide the compilers directly, just wraps the system install with some activation scripts. IMO a *metapackage* is a package that only lists other packages as dependencies, but doesn't deal with anything else like complex mechanisms to set environment variables.

Thus, I'd say something like this:


```suggestion
  Unfortunately, this is not possible with Conda-Forge's current infrastructure (``nvcc``, ``cudatoolkit``, etc) if there is no local CUDA Toolkit installation. In particular, the ``nvcc`` package provided on Conda-Forge is a *wrapper package* that exposes the actual ``nvcc`` compiler to our CI infrastructure in `conda`-friendly way; it does not contain the full ``nvcc`` compiler toolchain. One of the reasons is that CUDA headers like ``cuda.h``, ``cuda_runtime.h``, etc, are not redistributable according to NVIDIA's EULA, which are needed at compile time. Likewise, the ``cudatoolkit`` package only contains CUDA runtime libraries and not the compiler toolchain.
```

leofang: Sounds good, thanks Jaime!

jaimergp: Awesome, thanks for the work here!

jaimergp: Oops, and sorry about the single backtick thing. Too much markdown 😬

leofang: No worries. I don't know the syntax here either; simply following what's done above 😂

beckermr (approved):

beckermr:
@leofang I meant instructions that describe how to use nvcc locally after you install the toolkit.

leofang:
> I meant instructions that describe how to use nvcc locally after you install the toolkit.

@beckermr I added two links. Users can just run `nvcc` directly AFAIK, nothing really special. Do you expect a note like "you do not have to install `nvcc` from Conda-Forge; actually it might interfere with `nvcc` from CUDA Toolkit"?

beckermr:
> Users can just run nvcc directly AFAIK, nothing really special.

This is the information I was looking for! No need to add it. LGTM!

leofang:
Thanks everyone!

PR #1364, opened 2021-04-12
Minor fixes the Becoming Involved section

PrernaSingh587:
PR Checklist:


beckermr (changes requested):
We do not need to change these links. They will break this way if we ever reorganize the docs.

PR #1387, opened 2021-04-15
Update 00_intro.rst

bashirk:
PR Checklist:

Added the CRAN and R terms to glossary

### DIFF
--- a/src/misc/00_intro.rst
+++ b/src/misc/00_intro.rst
@@ -21,11 +21,8 @@ Glossary
   ICU
     **I**\ nternational **C**\ omponents for **U**\ nicode. ICU is an open-source project of mature C/C++ and Java libraries for Unicode support, software internationalization, and software globalization. `Learn More <http://site.icu-project.org/>`_. 
     
-  R
-    **R** is a free software environment for statistical computing and graphics. It compiles and runs on a wide variety of UNIX platforms, Windows and MacOS. `Learn More <https://r-project.org/>`_.
-
   CRAN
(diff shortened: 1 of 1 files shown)


src/misc/00_intro.rst
```diff
@@ -21,12 +21,16 @@ Glossary
   ICU
     **I**\ nternational **C**\ omponents for **U**\ nicode. ICU is an open-source project of mature C/C++ and Java libraries for Unicode support, software internationalization, and software globalization. `Learn More <http://site.icu-project.org/>`_. 
     
-    
+  R
```
viniciusdc, line 24: I don't see a reason for R being referred here, it's not something that appears too often in the docs and we can get the grasp of it by the context

src/misc/00_intro.rst
```diff
@@ -21,12 +21,16 @@ Glossary
...
+    **R** is a free software environment for statistical computing and graphics. It compiles and runs on a wide variety of UNIX platforms, Windows and MacOS. `Learn More <https://r-project.org/>`_.
+
+  CRAN
```
viniciusdc, line 27: This one, I really appreciate been added here, as I also had to look for this in the past.

viniciusdc (commented):
LGTM

src/misc/00_intro.rst
bashirk: Thanks for the feedback. I was thinking R would be a good addition since it is mentioned in the CRAN definition. Thanks again, I'll update accordingly.

Issue #1389, opened 2021-04-16
What is a recipe ?


Aaishpra:
- [x] I read [the conda-forge documentation](https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge) and could not find the solution for my problem there.

### Issue:
Throughout the conda-forge documentation, there is a mention of the word recipes, How to add a recipe, how to update a recipe,
But no mention about what is a recipe for packages, in my view there should be some sort of small introduction about what is a recipe and it can be added at the homepage.

Aaishpra:
@viniciusdc if this issue seems legit and needed I can work here or you can close the issue thanks

beckermr:
A recipe is the set of files needed to build the conda package. This includes at minimum the meta.yaml, but might also include license files, patches, build scripts, test scripts etc. If you look in any feedstock, the recipe is all of the stuff in the recipe directory.

Aaishpra:
Yes, but can that be included on homepage so someone visiting the website for the first time can directly know what it is.

beckermr:
Sure. I'd make a glossary entry and add some links

viniciusdc:
> Yes, but can that be included on homepage so someone visiting the website for the first time can directly know what it is.

Even though we don't have a straighforward directly definition of what a recipe is, it's conteints and uses are well written in the docs. There is still a full exemple of it (see more [here](https://conda-forge.org/docs/maintainer/adding_pkgs.html#the-recipe-meta-yaml)

Aaishpra:
Ya, I saw it and it perfectly explained me how a recipe can be created or what it's contents are, but due to the excessive use of the word recipe, i feel there should be some proper definition of this word and this definition should be present at the homepage.

viniciusdc:
> Ya, I saw it and it perfectly explained me how a recipe can be created or what it's contents are, but due to the excessive use of the word recipe, i feel there should be some proper definition of this word and this definition should be present at the homepage.

sure, I totally understand. I think the best answer to this was given by Matt above, also we could adapt the definition from [conda-recipes](https://docs.conda.io/projects/conda-build/en/latest/concepts/recipe.html) and use this chance to explain what the final structure of the feedstock will be... @beckermr any thoughts on this?

beckermr:
Yup. It is anything in the recipe folder in a feedstock. It always has a meta.yaml. It can have a bunch of other things.

Aaishpra:
[...]

PR #1395, opened 2021-04-20
Recipe definition added 

Aaishpra:
Recipe definition added to glossary and links in homepage added

PR Checklist:

image

Fixes: #1389 

### DIFF
--- a/src/misc/00_intro.rst
+++ b/src/misc/00_intro.rst
@@ -6,6 +6,9 @@ Glossary
 
 .. glossary::
 
+  Recipe
+    **R**\ ecipe is the collection of files required to build the conda package. This includes at minimum the `meta.yaml <https://conda-forge.org/docs/maintainer/adding_pkgs.html#the-recipe-meta-yaml>`_, but can also include license files, patches, build scripts, test scripts etc.
+  
   CI
     **C**\ ontinuous **I**\ ntegration.
 
--- a/src/misc/00_intro.rst
+++ b/src/misc/00_intro.rst
@@ -7,7 +7,8 @@ Glossary
 .. glossary::
 
   Recipe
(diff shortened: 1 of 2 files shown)


src/misc/00_intro.rst
```diff
@@ -6,6 +6,9 @@ Glossary
 
 .. glossary::
 
+  Recipe
+    **R**\ ecipe is the collection of files required to build the conda package. This includes at minimum the `meta.yaml <https://conda-forge.org/docs/maintainer/adding_pkgs.html#the-recipe-meta-yaml>`_, but can also include license files, patches, build scripts, test scripts etc.
```
ForgottenProgramme, line 10: Recipe is the collection of files required to build a (not the) conda package. The includes (comma) at minimum (comma) the meta.yaml file...
Maybe you can add a link to the sample recipe. (https://github.com/conda-forge/staged-recipes/tree/master/recipes/example)

Aaishpra: Done, thanks for suggesting

viniciusdc (commented):
I think this addition to the glossary is valid, but the link in the main page still looks kinda strange to me (though I liked it )

Aaishpra:
Basically we don't have separate links for every definition in the glossary, ( although i think there should be) so it felt strange to me also. But if you want we can remove the links from homepage.

viniciusdc:
> Basically we don't have separate links for every definition in the glossary, ( although i think there should be) so it felt strange to me also. But if you want we can remove the links from homepage.

@beckermr any thoughts on this? (My idea was set it and check if someone complains ;p)

Aaishpra:
@viniciusdc @beckermr  now as I think more, I think there is absolutely a need for separate links/references to different glossary descriptions so that they can be properly referenced in other places.
Something like [here](https://conda-forge.org/docs/maintainer/conda_forge_yml.html#azure), where we have different sub-headings as well as references for every config.
If you think this can be worked on, I would like to open an issue and work on the same.

beckermr:
I don't have a preference here.

jaimergp:
Integrated in #2087. Thanks!

PR #1442, opened 2021-05-14
Added faq about supported languages 

### LINKED ISSUE #986: List of supported languages
While I think we have mostly dispelled the notion of Conda being strictly a Python package manager, it would be helpful for people to have more visibility into what other languages are supported. Along these lines it might be nice to have a list of supported languages in the docs or perhaps the webpage somewhere.

Aaishpra:
Fixes #986

Added a faq about supported languages in conda. 

PR Checklist:

### DIFF
--- a/src/user/faq.rst
+++ b/src/user/faq.rst
@@ -7,6 +7,16 @@ FAQ
 
   We have an overview and step-by-step instruction on contributing packages in the section :ref:`dev_contribute_pkgs`.
 
+.. _faq_supported_lang:
+
+:ref:`(Q) <faq_supported_lang>` **What languages does conda-forge support?**
+
+  Following are some languages conda-forge supports.
+   - Python                  
+   - R	
+   - Fortran 
+   - C++ 	
+  Learn More about about packages per language `here <https://github.com/conda-forge/by-the-numbers/blob/master/packages_per_lang.ipynb>`_.
 .. _faq_pkg_update:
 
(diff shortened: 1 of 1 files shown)


Aaishpra:
@viniciusdc @beckermr please have a look here. I was not sure what could be the best place to add this information that's why mentioned it in a FAQ

viniciusdc:
> @viniciusdc @beckermr please have a look here. I was not sure what could be the best place to add this information that's why mentioned it in a FAQ

Uhm, I loved the link you posted there, but I don't think this should be in the FAQ section... A more appropriate place might be somewhere in the intro (I think we have an issue somewhere else with some indications). And we should change supported languages by something similar but not too strong, like: Currently conda-forge provides packages for an increasing number of languages such as .... (even the provides might be too strong here)

Also, that notebook you saw is currently inspecting the repo that contain all the (meta)data of the feedstocks, issues, PR's we have :smile:

Aaishpra:
> Uhm, I loved the link you posted there, but I don't think this should be in the FAQ section... A more appropriate place might be somewhere in the intro (I think we have an issue somewhere else with some indications). 

I was also thinking the same, can we create a new heading or just add it with the intro. And i think the issue you are talking about is #986

>And we should change supported languages by something similar but not too strong, like: Currently conda-forge provides packages for an increasing number of languages such as .... (even the provides might be too strong here)

Great, I got your point, after you confirm the place i will change it into a more lucid way.

> Also, that notebook you saw is currently inspecting the repo that contain all the (meta)data of the feedstocks, issues, PR's we have 😄

 It sounds something similar to an API for a web application,  nevertheless Thanks for the information👍🏻

PR #1549, opened 2021-11-11
Add favicon

BastianZim:
PR Checklist:

- [ ] make all edits to the docs in the `src` directory, not in `docs` or in the html files
- [x] note any issues closed by this PR with [closing keywords](https://help.github.com/articles/closing-issues-using-keywords)
- [x] put any other relevant information below

It's a bit difficult to test because not everything is built by the `src` script but I built the rest using the python scripts and it works. The only thing I couldn't test is `feedstock_outputs.html.tmpl` as I'm not sure how to.

PR #1550, opened 2021-11-11
Add favicon

BastianZim:
PR Checklist:

- [ ] make all edits to the docs in the `src` directory, not in `docs` or in the html files

Followup to https://github.com/conda-forge/conda-forge.github.io/pull/1549

Copying my comment from there:

> Everything's working except https://conda-forge.org/feedstock-outputs/ which is (I think) because feedstock_outputs.html.tmpl is actually not used anymore: https://github.com/conda-forge/conda-forge.github.io/search?q=feedstock_outputs.html.tmpl
>
> Only feedstocks.html.tmpl is used (Which I didn't change because the link is feedstock-outputs/).
>
> Is this assumption correct?

If yes, might make sense to delete that.

### DIFF
--- a/feedstocks.html.tmpl
+++ b/feedstocks.html.tmpl
@@ -9,6 +9,8 @@
     <meta name="description" content="">
     <meta name="author" content="">
 
+    <link rel="shortcut icon" href="img/favicon.ico">
+
     <title>conda-forge feedstocks | community driven packaging for conda</title>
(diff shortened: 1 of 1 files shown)


beckermr:
This file is used. We don't use the repo anymore.

BastianZim:
Ahh ok.

PR #1556, opened 2021-11-18
Improve build_platform and providers documentation

carterbox:
Clarify that build_platforms are native when unspecified and that cross-compilation must be manually specified if (as in the case of osx_arm64 right now), there is no native provider available.

Closes conda-forge/conda-smithy#1546

PR Checklist:

### DIFF
--- a/src/maintainer/conda_forge_yml.rst
+++ b/src/maintainer/conda_forge_yml.rst
@@ -129,14 +129,26 @@ automatic version updates/migrations for feedstocks. The current options are
 build_platform
 --------------
 This is a mapping from the build platform to the host platform of the package
-to be built. For eg: following builds a ``osx-64`` platform from ``linux-64``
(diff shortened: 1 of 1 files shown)


src/maintainer/conda_forge_yml.rst
```diff
@@ -378,6 +392,8 @@ To enable ``linux_ppc64le`` and ``linux_aarch64`` and the following:
       linux_ppc64le: default
       linux_aarch64: default
 
+If the ``build_platform`` for an arch is not available with the selected provider, the build will be disabled; cross-compilation must be specified manually.
```
carterbox, line 395: I'm actually not sure if the build will be disabled or if the build will be created but just probably fail.

beckermr: We emulate aarch64.

beckermr: What default means is actually set in smithy itself and subject to change.

PR #1562, opened 2021-11-25
Under the hood docs improvements

croth1:
Things that happened:

- move from recommonmark to myst and lift sphinx version restrictions
- fix all remaining sphinx compiler warnings
- changed the sphinx compile flags such that the build errors out when there are warnings.
- check validity of links automatically on build.

Fingers crossed that this will work 😅


croth1:
Ah grr, I just saw that one of my sed commands went rogue. I will need to do some postprocessing 👎

Issue #1563, opened 2021-11-25
Potentially broken links


croth1:
Here is a list of potentially broken external links in the docs (excluding the minutes). Some of the forbidden ones seem ok, but the 404 definitely need fixing.

List is generated by `make linkcheck` in the src folder.

- [x] (maintainer/maintainer_faq: line   70) broken    https://anaconda.org - 403 Client Error: Forbidden for url: https://anaconda.org/
- [x] (user/introduction: line   16) broken    https://anaconda.org/ - 403 Client Error: Forbidden for url: https://anaconda.org/
- [x] (maintainer/maintainer_faq: line   70) broken    https://cloudflare.com/learning/cdn/what-is-a-cdn/ - 403 Client Error: Forbidden for url: https://www.cloudflare.com/learning/cdn/what-is-a-cdn/
- [x] (maintainer/infrastructure: line  260) broken    https://circleci.com/docs/getting-started/#add-and-follow-more-projects - 404 Client Error: Not Found for url: https://circleci.com/docs/getting-started/
- [x] (maintainer/infrastructure: line  260) broken    https://circleci.com/add-projects - 404 Client Error: Not Found for url: https://circleci.com/add-projects
- [x] (   misc/00_intro: line   16) broken    https://conda-forge.org/docs/maintainer/knowledge_base.html#core-dependency-tree-packages-cdts/ - Anchor 'core-dependency-tree-packages-cdts/' not found
- [x] (maintainer/infrastructure: line  274) broken    https://conda-forge.org/docs/webservice.html#conda-forge-admin-please-update-circle - 404 Client Error: Not Found for url: https://conda-forge.org/docs/webservice.html
- [x] (maintainer/knowledge_base: line 1377) broken    https://conda-forge.org/status/#armosxaddition - Anchor 'armosxaddition' not found
- [x] (maintainer/maintainer_faq: line  160) broken    https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#inherited-environment-variabless - Anchor 'inherited-environment-variabless' not found
- [x] (   misc/00_intro: line   10) broken    https://en.wikipedia.org/wiki/Continuous_integration/ - 404 Client Error: Not Found for url: https://en.wikipedia.org/wiki/Continuous_integration/
[...]

PR #1565, opened 2021-11-26
[WIP] Clean up broken bits and links in the docs

croth1:
I will continuously fix small doc issues here, please feel free to comment, but do not merge just yet 😄 

- [x] remove references to CircleCI from the docs.


jaimergp:
I have a used 404 checkers in CI in the past. Maybe it's a valuable addition here!

For example:

* https://github.com/marketplace/actions/404-links
* https://github.com/marketplace/actions/broken-link-check
* https://github.com/JustinBeckwith/linkinator-action

And I am sure there are others.

croth1:
Yea, I am thinking about doing the checks automatically - will have a look into your suggestions, thanks. So far I used `linkcheck` directly from sphinx. It works, but I see two issues/downsides:

1. it also reports broken links in the meeting minutes, which I don't think are worth maintaining. I will need to find a way to prevent that.
2. there are a lot of github links and github's API seems to throttle the number of requests we can send. It might not be fast enough for checking for every commit.

jaimergp:
I have asked my colleagues for their experience in this task, and @melissawm recommends the following:

> * Sphinx catches some dead links with -n (nitpicky mode) https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-n
> * I recently also learned about Lychee: https://github.com/lycheeverse/lychee
> * This has been configured in napari if you want to see how it looks like in action: https://github.com/napari/napari.github.io/blob/main/.github/workflows/link-checker.yml

beckermr (changes requested):
Why did you remove references to circle? We still support it even if it is not used very much.

croth1:
Hmh, oh I assumed it's not longer supported. I can keep it, but then I have to figure out how to fix the broken links to the CircleCI configuration 😬

croth1:
So should be done now. Closes #1563.

beckermr:
thank you!

Issue #889, opened 2019-10-04
Meta Docs!


CJ-Wright:
It would be nice to add some docs to the Readme and/or docs about how to edit the docs. This can be particularly confusing, since we have multiple dirs which have docs, but only one dir that will actually impact the docs, `src`.

xref: https://github.com/conda-forge/conda-forge.github.io/issues/884 and a whole bunch of my doc PRs 🐑

jakirkham:
Maybe a contribution guide and PR checklist would be good

sodre:
Should we switch to using the gh-pages branch for hosting the published website? That would reduce the confusion.

jakirkham:
Is that possible on the webpage repo? Thought we were limited to `master`

sodre:
@jakirkham, you're right. I fixed my comment.

jakirkham:
No worries. Agree that would be easier if possible 🙂

ForgottenProgramme:
I guess, this has been fixed. 
PR Checklist has been created. 
Also, the 'improving the documentation' section (https://conda-forge.org/docs/user/contributing.html#improve-the-documentation) mentions the following:

>The source text is stored there in the src/ subdirectory and is formatted using Python’s reStructuredText system.

Perhaps, to clarify further, after '..stored in the src/ subdirectory' we can add 'and NOT in the /docs subdirectory'. 

@jakirkham 
@CJ-Wright

kathatherine:
@beckermr This able to be closed? I don't have permissions myself.

beckermr:
thank you!

PR #1617, opened 2022-02-24
Added clarifying statement to Contributing page.

kathatherine:
Hey, all, just looking to add the clarification Mahe spoke about in #889, so that we could get that issue closed and not clogging up the issue tracker for our 2022 Outreachy applicants.


beckermr:
Thank you!

Issue #1370, opened 2021-04-12
conda-forge , Conda-forge or Conda-Forge? 


ForgottenProgramme:
At most(80%, I'd guess) places in the conda-forge docs, 'conda-forge' is written with all characters in lowercase, even if it comes at the beginning of the sentence. 

But at some places this style(rule?) is not maintained. 

Infact, I'm seeing PRs changing 'conda-forge' to 'Conda-forge' or 'Conda-Forge'.

This might not seem very important, but a lack of consistency of style would reflect badly on us. 

It is hence required that we clarify what would be the 'style' we'd be following.

ForgottenProgramme:
@leofang @beckermr @viniciusdc @ocefpaf

beckermr:
I'd say always lower-case and in double backticks, but that is me.

ForgottenProgramme:
> I'd say always lower-case and in double backticks, but that is me.

I second that.

beckermr:
Other members of @conda-forge/core should comment too.

leofang:
I suppose it also depends on if it is referred to as a brand/org/service/group of people or as a (Conda) channel.

ForgottenProgramme:
> I suppose it also depends on if it is referred to as a brand/org/service/group of people or as a (Conda) channel.

oh, that might be a good point! 
When referred to as a channel, all lowercase and within double backticks is the right away. Because that's how we write other channel names too. 

Brand, organization, service; these three are not very far from one another, I guess?

bashirk:
> At most(80%, I'd guess) places in the conda-forge docs, 'conda-forge' is written with all characters in lowercase, even if it comes at the beginning of the sentence.
> 
> But at some places this style(rule?) is not maintained.
> 
> Infact, I'm seeing PRs changing 'conda-forge' to 'Conda-forge' or 'Conda-Forge'.
> 
> This might not seem very important, but a lack of consistency of style would reflect badly on us.
> 
> It is hence required that we clarify what would be the 'style' we'd be following.

Great observation. I would want to believe ``conda-forge`` is the better naming style from the lots, and it has been the most consistent.

ForgottenProgramme:
>I would want to believe conda-forge is the better naming style from the lots, and it has been the most consistent.

@bashirk , true that!

PR #1642, opened 2022-03-25
Formatting docs - backticks and lowercase for 'conda-forge'

chicken-biryani:
This PR aims at fixing #1370

This PR tries to fix the inconsistency across docs by changing mentions of  Conda-forge or Conda-Forge to ``conda-forge`` (Double backticks)

PR Checklist:

- [X] make all edits to the docs in the `src` directory, not in `docs` or in the html files
- [X] note any issues closed by this PR with [closing keywords](https://help.github.com/articles/closing-issues-using-keywords)


chicken-biryani:
Hi @beckermr and @kathatherine can you please take a look and let me know if I could better the PR?

(Also, thank god for VS Code 😅🤞)

beckermr:
Looks great!

jaimergp:
I wonder if at this point we should consider RST [substitutions](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/reuse.html#use-a-substitution). That way we always use `|conda-forge|` and standardize in one place how this is rendered. But if this complicates the discussion, we can also leave it aside.

chicken-biryani:
@jaimergp that's okay. If it will help in the wrong run, I'll change it.

Maybe I can experiment with a few lines first. If it doesn't work out I'll drop the commit.

hmaarrfk:
@jaimergp i find that the substitutions increase indirection and make it really difficult to read the source text.

I would personally prefer adding a linting step to catch poorly formatted words in the RST.

jaimergp:
One problem I see with always using `conda-forge` instead of conda-forge is that in RST is not possible to have monospace words in a hyperlink, which would lead to inconsistencies. As a result I am inclined to say we should just have always lowercased, unless it starts a sentence, much like what `conda` suggests in their [CEP README](https://github.com/conda-incubator/ceps#conda-capitalization-standards).

src/maintainer/infrastructure.rst
```diff
@@ -343,7 +343,7 @@ works as follows.
    and some information about the package it is trying to upload.
 3. The webservices server validates the secret token, the integrity of the package, and
    that the package is allowed for the given feedstock.
-4. If all of the validation passes, the package is then copied to the ``conda-forge``
+4. If all of the validation passes, the package is then copied to the conda-forge
```
jaimergp, line 346:
```suggestion
4. If all of the validation passes, the package is then copied to the ``conda-forge``
```

kathatherine (approved):
The limitations of rst links make this the best solution. I agree, @jaimergp, and it works for me.

PR #1716, opened 2022-04-14
Fix #1711: Add Cheat sheet to User documentation

### LINKED ISSUE #1711: Add a Cheat sheet to the User documentation
### Where should the content be added?

Tips and tricks section

### What should be added?

Include  a Cheatsheet with frequently used commands.  

### Additional information

It would improve a new user's experience if there is a section to reference for most frequently used commands. Commands for creating conda environments, listing conda environments, switching between environments, installing packages.

Hil-dred:
Hello @beckermr @kathatherine, kindly review my PR. I have created this PR as a potential solution to #1711

### DIFF
--- a/src/user/tipsandtricks.rst
+++ b/src/user/tipsandtricks.rst
@@ -1,3 +1,38 @@
+Cheat sheet
+===========
+
+Create a new conda environment
+------------------------------
+
+.. code-block:: shell
+
+    $ conda create --name ENVIRONMENT_NAME
+
+
+Activate new conda environment
+------------------------------
+
+.. code-block:: shell
+
+    $ conda activate ENVIRONMENT_NAME
+
+
+List conda environments
+-----------------------
+
+.. code-block:: shell
+
+    $ conda info --envs
+
+
+Install packages
+-----------------
+
+.. code-block:: shell
+
+    $ conda install PACKAGE_NAME
+
+
 Tips & tricks
 *************
 


kathatherine (approved):
Looks great, thanks!

jakirkham:
Wonder if we want this in the conda-forge or Conda docs. No strong feelings either way. This has just come up in the past with documentation like this

Hil-dred:
Hello @jakirkham Out of curiosity, may I know why this may not be a preferred addition?

jakirkham:
Previously when more Conda specific documentation has come up, the suggestion has been to include the content there. The thinking being we don't want to have duplicate information that may go stale and we would want there to be one place for users to naturally look for this information

If we were wanting to reference Conda docs, [this page]( https://docs.conda.io/projects/conda/en/latest/commands.html ) looks similar to what we might want. Though maybe there are other ways one might want to update Conda's docs given the ideas here

No strong feelings on either approach

kathatherine:
Those conda command references are created automatically from the CLI help, so they will definitely remain more up to date overall. That's a good point about redundancy. There's also a basic conda cheatsheet that was recently updated and linked [here](https://docs.conda.io/projects/conda/en/latest/user-guide/cheatsheet.html).

Hil-dred:
@jakirkham ,Thanks for explaining 👍

PR #1795, opened 2022-07-06
Remove obsolete documentation

jens-ox:
This PR removes the `docs` folder, as it seems to be autogenerated by GitHub Actions. Additionally, the obsolete pages `search.html` and `artifact.html` have been removed.

`elm` seemed to be only used on those two pages, so that was also removed.


beckermr (changes requested):
This PR removes key instructions and CI workflows and so cannot be merged.

jens-ox:
@beckermr which key instructions/CI workflows? The `docs-linter` workflow isn't needed anymore (as there is no more `docs` folder) and the instructions regarding putting changes in `src` instead of `docs` is also obsolete.

beckermr:
The website is run live from the docs folder. We cannot delete it.

jens-ox:
Not sure if I understand the code completely wrong, but it seems like the `docs` folder [even gets deleted by the action](https://github.com/conda-forge/conda-forge.github.io/blob/main/.ci_scripts/update_docs#L24) prior to re-generating and deploying it.

beckermr:
Ahhhhh my understanding of this repo is likely out of date. Sorry for the trouble! Let me take another look.

beckermr:
OK. I think this PR is safe to merge. I am going to squash it to one commit so we can back it out more easily later if we need to.

jaimergp:
Should we also add `docs/` to `.gitignore` now? There's a risk existing forks/clones re-push that folder in a PR, iiuc.

PR #1799, opened 2022-07-16
Suggest that people add packages in a random location

hmaarrfk:
I often review these requests and when two come in at the same time, this suggestion causes conflict.

I'm hoping this can cause a smoothly workflow.


beckermr (approved):

hmaarrfk:
Thanks.

PR #1821, opened 2022-08-29
Switch # artifacts to millions

jaimergp:
PR Checklist:

conda-forge is now over 1M artifacts! That means conda-forge.org renders it as 1011k artifacts... and 1M looks cooler :D

### DIFF
--- a/js/stats.js
+++ b/js/stats.js
@@ -34,7 +34,7 @@ function displayStatsJSON (reportText) {
   div.innerHTML = (
     (stats.n_repos / 1e3).toFixed(1) + 'k feedstocks<br>' +
     (stats.n_packages / 1e3).toFixed(1) + 'k packages<br>' +
-    (stats.n_artifacts / 1e3).toFixed(0) + 'k artifacts'
+    (stats.n_artifacts / 1e6).toFixed(1) + 'M artifacts'
   )
 
   div = document.getElementById('stats-issues-prs')


js/stats.js
```diff
@@ -34,7 +34,7 @@ function displayStatsJSON (reportText) {
   div.innerHTML = (
     (stats.n_repos / 1e3).toFixed(1) + 'k feedstocks<br>' +
     (stats.n_packages / 1e3).toFixed(1) + 'k packages<br>' +
```
jakirkham, line 36: Should we update these too?

js/stats.js
```diff
@@ -34,7 +34,7 @@ function displayStatsJSON (reportText) {
...
-    (stats.n_artifacts / 1e3).toFixed(0) + 'k artifacts'
+    (stats.n_artifacts / 1e6).toFixed(1) + 'M artifacts'
```
jakirkham, line 37: Is there a way to automate bumping this? Perhaps some JavaScript package that renders numbers in scientific format or similar?

jakirkham (commented):
Thanks Jaime! 🙏

Had a couple follow up questions below

js/stats.js
jaimergp: This is how it looks like now:

image


I don't anticipate another change anytime soon, but I guess we could have a little function to automate the formatting.

jaimergp: e.g. like this https://stackoverflow.com/a/60550008/3407590

PR #1853, opened 2022-11-17
Fix `run_constrained` typo in `adding_pkgs.rst`

jrbourbeau:
It looks like `run_dependencies` should be replaced with `run_constrained` here


jakirkham (approved):
Thanks James! 🙏

Issue #1732, opened 2022-04-22
Move to vc142 toolchain on windows


h-vetinari:
## TL;DR

Move to newer toolset because:
* vc141 is disappearing in the CI of many projects (due to the compiler where it's the default being EOL), leading to bitrot, breakage, and extra work (both upstream and in packaging)
* ABI not hard-linked with OS as on unix
* better performance & less bugs
* it's trivial for conda-forge to keep a consistent toolchain
* the set of users being broken by this change is either empty or very small (<- happy to be challenged on this)

### Visual Studio version vs. Toolset version

The MSVC compiler comes in different versions (..., VS2015, VS2017, VS2019, VS2022); these compilers come with a default toolset, but are able to target older ones. E.g. VS2017 comes with vc141, but can target vc140 (the default toolset of VS2015). vc142 is the default toolset of VS2019. What is the toolset exactly? From the [docs](https://docs.microsoft.com/en-us/cpp/build/projects-and-build-systems-cpp?view=msvc-170#the-msvc-toolset):
> The Microsoft C++ compiler, linker, standard libraries, and related utilities make up the MSVC compiler toolset.

### ABI compatibility

Microsoft has been _extremely_ conservative with ABI-compatibility from VS2015-VS2022 (the story about `[[msvc::no_unique_address]]` is instructive for that), meaning that a given project can use the new compilers while using all the old binaries that flow into the build. The one [caveat](https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-170) is that any library built with a given toolset version can only be consumed (e.g. linked) by a toolset version that's the same or newer.

### No OS coupling

Furthermore, unlike on linux, the version / ABI of the compiler is not coupled as tightly to a distribution (mostly through the ABI of the libstdcxx that comes with the compiler). As such, changing the toolchain version does not affect which platforms a binary built by a given toolset will run on.

### VS2017 EOL & impact on open-source CI

VS2017 has [reached](https://docs.microsoft.com/en-us/lifecycle/products/visual-studio-2017) EOL, and has been aggressively removed by microsoft from its public CIs (Azure & GHA), see previous [discussion](https://github.com/conda-forge/conda-forge.github.io/issues/1538). This is IMO both understandable (avoid maintaining 3 separate compilers), as well as not very impactful, due to the very strong ABI-compatibility between VS2017 and VS2019.

[...]

PR #1956, opened 2023-05-10
Update maintainer kb windows doc

mfisher87:
PR Checklist:

I went through the process of testing on Windows today, and as a non-Windows user I felt these changes would have helped me! A little bit of information was out-of-date.

### DIFF
--- a/src/maintainer/knowledge_base.rst
+++ b/src/maintainer/knowledge_base.rst
@@ -133,15 +133,12 @@ a workaround for now.
 Testing a local build
 ^^^^^^^^^^^^^^^^^^^^^
 
-Using a local channel looks a bit different in Windows, too. The quotes are
-important:
+Because we're using ``conda-build`` directly instead of ``build-locally.py``,
+we can use the ``local`` channel:
 
 .. code-block::
 
     conda create -n my-new-env -c local my-package
-
-If you installed from a different distribution than ``miniforge`` or to a
-different path, you'll need to alter this path.
 
 
 Notes on native code
--- a/src/maintainer/knowledge_base.rst
+++ b/src/maintainer/knowledge_base.rst
@@ -156,7 +156,7 @@ While you can obtain these tools by installing the right version of the full
 `Visual Studio <https://visualstudio.microsoft.com/>`__ development
 environment, you can save a lot of time and bandwidth by installing standalone
 “build tools” packages. You can get them from `Visual Studio
-Subscriptions<https://visualstudio.microsoft.com/vs/older-downloads/#visual-studio-2019-and-other-products>`__.
+Subscriptions <https://visualstudio.microsoft.com/vs/older-downloads/#visual-studio-2019-and-other-products>`__.
 To download build tools, you'll need a Microsoft account. Once on the
 Visual Studio Subscriptions page, you may also need to join the Dev Essentials
 program. Once that's done, you can click the "Download" tab and search for
--- a/src/maintainer/knowledge_base.rst
+++ b/src/maintainer/knowledge_base.rst
@@ -242,9 +242,9 @@ In ``recipe/conda_build_config.yaml`` file:
 .. code-block:: yaml
 
     c_compiler:    # [win]
-    - vs2022       # [win and x86_64]
+    - vs2022       # [win]
     cxx_compiler:  # [win]
-    - vs2022       # [win and x86_64]
+    - vs2022       # [win]
 
 You can look at the changes in `this PR <https://github.com/conda-forge/vcpkg-tool-feedstock/pull/41/files>`__.
 
--- a/src/maintainer/knowledge_base.rst
+++ b/src/maintainer/knowledge_base.rst
@@ -119,7 +119,7 @@ scope of this documentation, there are ample discussions on them on the broader
 Internet.
 
 To bootstrap a conda environment and install ``conda-build``, consider
-`miniforge<https://github.com/conda-forge/miniforge>`__.
+`miniforge <https://github.com/conda-forge/miniforge>`__.
 
 
 Executing a build
--- a/src/maintainer/knowledge_base.rst
+++ b/src/maintainer/knowledge_base.rst
@@ -125,14 +125,14 @@ compiled libraries used in Python, `as described on this Python wiki page
 <https://wiki.python.org/moin/WindowsCompilers>`__. The current relevant
 versions are:
 
-* For Python 3.5–3.7: Visual C++ 14.0
+* For Python 3.5–3.10: Visual C++ 14.0
 
 While you can obtain these tools by installing the right version of the full
 `Visual Studio <https://visualstudio.microsoft.com/>`__ development
 environment, you can save a lot of time and bandwidth by installing standalone
 “build tools” packages. The links are as follows:
 
-* For Python 3.5–3.7: `Microsoft Build Tools for Visual Studio 2017
+* For Python 3.5–3.10: `Microsoft Build Tools for Visual Studio 2017
   <https://visualstudio.microsoft.com/vs/older-downloads/#visual-studio-2017-and-other-products>`__.
 
 If you need more information. Please refer `the Python wiki page on Windows compilers
--- a/src/maintainer/knowledge_base.rst
+++ b/src/maintainer/knowledge_base.rst
@@ -118,6 +118,30 @@ use a general web search to explore — while these topics are beyond the
 scope of this documentation, there are ample discussions on them on the broader
 Internet.
 
+To bootstrap a conda environment and install ``conda-build``, consider
+`miniforge<https://github.com/conda-forge/miniforge>`__.
+
+
+Executing a build
+^^^^^^^^^^^^^^^^^
+
+The ``build-locally.py`` script does not support Windows. Use ``conda build
+recipe/``.
+
+Using a local channel looks a bit different in Windows, too (note the
+triple-forward-slash in the channel name):
+
+.. code-block::
+
+    conda create -n my-new-env -c file:///C:\Users\User\miniforge3\conda-build my-package
+
+If you installed from a different distribution than ``miniforge`` or to a
+different path, you'll need to alter this path.
+
+
+Notes on native code
+--------------------
+
 In order to compile native code (C, C++, etc.) on Windows, you will need to
 install Microsoft’s Visual C++ build tools on your VM. You must install
 particular versions of these tools — this is to maintain compatibility between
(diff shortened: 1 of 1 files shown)


src/maintainer/knowledge_base.rst
```diff
@@ -118,28 +118,55 @@ use a general web search to explore — while these topics are beyond the
...
+
+.. code-block::
+
+    conda create -n my-new-env -c file:///C:\Users\User\miniforge3\conda-build my-package
+
+If you installed from a different distribution than ``miniforge`` or to a
+different path, you'll need to alter this path.
+
+
+Notes on native code
```
mfisher87, line 142: I didn't have to deal with native code to build my package today. I was distracted by the many sections on dealing with native code, so I felt they'd be better off together in one section.

src/maintainer/knowledge_base.rst
```diff
@@ -118,28 +118,55 @@ use a general web search to explore — while these topics are beyond the
 scope of this documentation, there are ample discussions on them on the broader
 Internet.
 
+To bootstrap a conda environment and install ``conda-build``, consider
+`miniforge<https://github.com/conda-forge/miniforge>`__.
```
jaimergp, line 122:
```suggestion
`miniforge <https://github.com/conda-forge/miniforge>`__.
```

src/maintainer/knowledge_base.rst
```diff
@@ -118,28 +118,55 @@ use a general web search to explore — while these topics are beyond the
...
+
+
+Executing a build
+^^^^^^^^^^^^^^^^^
+
+The ``build-locally.py`` script does not support Windows. Use ``conda build
+recipe/``.
```
jaimergp, line 129:
```suggestion
The ``build-locally.py`` script does not support Windows (yet, PRs welcome!).
You can use ``conda build recipe/ -m .ci_support/choose_your_config.yaml`` as
a workaround for now.
```

I think this is a bit more accurate!

src/maintainer/knowledge_base.rst
```diff
@@ -118,28 +118,55 @@ use a general web search to explore — while these topics are beyond the
...
+
 In order to compile native code (C, C++, etc.) on Windows, you will need to
 install Microsoft’s Visual C++ build tools on your VM. You must install
 particular versions of these tools — this is to maintain compatibility between
 compiled libraries used in Python, `as described on this Python wiki page
 <https://wiki.python.org/moin/WindowsCompilers>`__. The current relevant
 versions are:
 
-* For Python 3.5–3.7: Visual C++ 14.0
+* For Python 3.5–3.10: Visual C++ 14.x
```
jaimergp, line 152: Interesting how the Python wiki doesn't mention Python 3.11 yet...

mfisher87: Indeed. I was treating that page as the source of truth; do you know if Visual C++ 14.x actually supports 3.11?  It's not an area I know much about, and my web search didn't turn up an answer.

src/maintainer/knowledge_base.rst
mfisher87: This was the first "Anonymous hyperlink mismatch" and just pushed another commit that fixes the other. Woops!

jaimergp (approved):
LGTM!

I'll leave it open for the weekend in case some else from @conda-forge/core wants to add something (e.g. wrt the Python 3.11 / Visual C++ version topic). Maybe @h-vetinari knows something too? 🙏 

Otherwise I'll merge on Monday!

jaimergp (approved):
LGTM!

I'll leave it open for the weekend in case some else from @conda-forge/core wants to add something (e.g. wrt the Python 3.11 / Visual C++ version topic). Maybe @h-vetinari knows something too? 🙏 

Otherwise I'll merge on Monday!

src/maintainer/knowledge_base.rst
```diff
@@ -118,28 +118,56 @@ use a general web search to explore — while these topics are beyond the
...
+The ``build-locally.py`` script does not support Windows (yet, PRs welcome!).
+You can use ``conda build recipe/ -m .ci_support/choose_your_config.yaml`` as
+a workaround for now.
+
+Using a local channel looks a bit different in Windows, too (note the
+triple-forward-slash in the channel name):
+
+.. code-block::
+
+    conda create -n my-new-env -c file:///C:\Users\User\miniforge3\conda-build my-package
```
h-vetinari, line 137: I've had no problems using unmodified local paths as channels if they're escaped by quotes, i.e.
```
conda create -n my-new-env -c "C:\Users\me\path\to\wherever\I\unpacked\an\artefact"
```
I think this is easier than triple slashes resp. mixing between forward/backward slashes?

src/maintainer/knowledge_base.rst
```diff
@@ -203,7 +231,7 @@ To skip building with a particular ``vc`` version, add a skip statement.
         - {{ compiler('cxx') }}
 
 Using vs2019
--------------
+^^^^^^^^^^^^^
```
h-vetinari, line 234: Since you're touching this section, I'll note that it's outdated and should be updated; https://github.com/conda-forge/conda-forge.github.io/pull/1878 has been open for a while, I guess it could be folded into this PR?

h-vetinari (commented):
> Maybe @h-vetinari knows something too? 🙏

Thanks for the ping, left some small comments/observations.

src/maintainer/knowledge_base.rst
mfisher87: Thanks, that is much easier :P I don't know Windows well so I struggled a bit with this when testing on a VM. Will update shortly...

mfisher87: 60c06a8f

src/maintainer/knowledge_base.rst
mfisher87: My best shot at that: e99d98c2

src/maintainer/knowledge_base.rst
```diff
@@ -234,26 +234,32 @@ To skip building with a particular ``vc`` version, add a skip statement.
...
-<https://github.com/conda-forge/libignition-msgs1-feedstock/pull/73/commits/81b5ee0e1d23f7f20427dd80d04cf1f7321b441d>`__ commit.
+.. code-block:: yaml
 
-After making these changes don't forget to rerender with ``conda-smithy`` (to rerender manually use ``conda smithy rerender`` from the command line).
+    azure:
+      settings_win:
+        pool:
+          vmImage: windows-2022
 
+You can look at the changes in `this PR <https://github.com/conda-forge/vcpkg-tool-feedstock/pull/41/files>`__.
```
h-vetinari: Since https://github.com/conda-forge/conda-smithy/pull/1720 got merged, this whole section can be removed (it's implied in the comments & cross-references from #1878, but perhaps it wasn't obvious).

mfisher87: Ah. I saw the comments about the idea to change the default in `conda-smithy`, but did not see the linked and merged PR from Feb.

mfisher87: 21319730

I left the PR link in, but I added a comment to that PR to point out the section in `conda-forge.yml` can now be omitted.

src/maintainer/knowledge_base.rst
```diff
@@ -202,27 +234,22 @@ To skip building with a particular ``vc`` version, add a skip statement.
...
 
 .. code-block:: yaml
 
     c_compiler:    # [win]
-    - vs2019       # [win]
+    - vs2022       # [win]
     cxx_compiler:  # [win]
-    - vs2019       # [win]
-
+    - vs2022       # [win]
```
jaimergp, line 247: We need to account for the [new `arm64` architecture in Windows](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/360bec574e3c93156269d6475b2c5c46be57e304/recipe/conda_build_config.yaml#L4-L5), so the selector needs an update.

src/maintainer/knowledge_base.rst
```diff
@@ -202,27 +234,22 @@ To skip building with a particular ``vc`` version, add a skip statement.
...
+^^^^^^^^^^^^
 
-In ``conda_build_config.yaml`` file:
+In ``recipe/conda_build_config.yaml`` file:
 
 .. code-block:: yaml
 
     c_compiler:    # [win]
-    - vs2019       # [win]
+    - vs2022       # [win]
```
jaimergp, line 245:
```suggestion
    - vs2022       # [win and x86_64]
```

src/maintainer/knowledge_base.rst
jaimergp, line 247:
```suggestion
    - vs2022       # [win and x86_64]
```

src/maintainer/knowledge_base.rst
jaimergp: This has been updated now in the wiki and they do mention 3.12+, so:

```suggestion
* For Python 3.5–3.12+: Visual C++ 14.x
```

jaimergp:
Thanks for the updates @mfisher87! A couple more comments and we are good to go from my side!

mfisher87:
Thanks @jaimergp! I think I got everything :)

mfisher87:
>Maybe conda-forge will upgrade to v2022 generally in the future, but for now that's the status quo and that's what we should document.

I'm even more confused :) I'm having a slow-brain morning I guess! How does that affect this PR which is changing the docs from recommending vs2019 to vs2022?

src/maintainer/knowledge_base.rst
```diff
@@ -118,28 +118,60 @@ use a general web search to explore — while these topics are beyond the
...
+
+Testing a local build
+^^^^^^^^^^^^^^^^^^^^^
+
+Using a local channel looks a bit different in Windows, too. The quotes are
+important:
+
+.. code-block::
+
+    conda create -n my-new-env -c "C:\Users\User\miniforge3\conda-build" my-package
```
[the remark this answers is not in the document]
mfisher87: Good catch. I believe @msarahan had already recommended that to me, but I forgot to try that in Windows. `-c local` has worked great for me on Linux, but I haven't tried it yet on Windows. Checking to see if my VM still works and will get back to you :)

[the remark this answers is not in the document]
mfisher87: 92dd9176

src/maintainer/knowledge_base.rst
h-vetinari: I disagree here. The current pinning _has_ to split x86_64 and arm64 by necessity, but we don't. In particular, specifying `vs2022  # [win]` does nothing on the arm64-side.

I would argue that it's _better_ for a feedstock to be using a single compiler version per platform whenever possible[^1], and certainly the docs should reflect that. It's also plain simpler to read & understand.

[^1]: but the introduction of a new platform forces us to accept that, at least on conda-forge level

h-vetinari:
> > I don't really understand why we'd want to compile with `vs2019` for `x86`. Why not `vs2022` for both `arm64` and `x86_64` platforms? I know very little about the Windows world :)
> 
> I don't know much either 😬 Maybe conda-forge will upgrade to v2022 generally in the future, but for now that's the status quo and that's what we should document.

On windows, the compilers come as an integrated toolchain (not separate components like on unix). Updating the toolchain requires everyone who wants to compile against conda-forge windows artefacts to also upgrade their toolchain. It's a bit hard to quantify because it doesn't affect users just downloading conda packages, but we've been conservative about this (FWIW, I wanted to upgrade faster from vs2017 to vs2019, but there was opposition. Eventually it lead to https://github.com/conda-forge/conda-forge.github.io/issues/1732, which explains the situation more in-depth).

That's the baseline. However, win-arm64 is (effectively) a brand-new platform, so there's no-one relying on conda-forge win-arm64 artefacts of any kind yet, and we need the newer toolchain because arm64-support is still being ramped up (and vs2019 likely not good enough). 

So now we have to split the toolchain version per architecture, until such time when all of conda-forge moves to vs2022, then they can move in sync from then on.

I left a change [request](https://github.com/conda-forge/conda-forge.github.io/pull/1956#discussion_r1239033099) related to this.

src/maintainer/knowledge_base.rst
mfisher87: Sounds like we should revert 7cc08a7?

h-vetinari: Yes please

mfisher87: 7af43c12

src/maintainer/knowledge_base.rst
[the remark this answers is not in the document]
mfisher87: Woopsie, forgot to also delete some prose that was impacted by the change to the code block. 30150602

jaimergp:
Thanks @h-vetinari! I think this is good to go now then!

jaimergp (approved):

h-vetinari (approved):
There's still some heavily outdated stuff in the "Building for different VC versions" section, but we can tackle that in a separate PR IMO, because it might need more work/discussion, and I don't want to hold up this PR.

Thanks for the PR & the patience!

mfisher87:
>Thanks for the PR & the patience!

Any time, thanks for being welcoming :)

PR #1980, opened 2023-07-13
ENH set date for cos6 EOL

beckermr:
PR Checklist:

### DIFF
--- a/src/user/announcements.rst
+++ b/src/user/announcements.rst
@@ -8,20 +8,31 @@ Our announcements are published to an RSS feed `here <https://conda-forge.org/do
 2023
 ----
 
+2023-07-12: EOL for CentOS 6
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+As you may be aware, we have delayed the deprecation of our CentOS 6 build
+system the ``linux64`` platform several times. We have now set a formal deprecation
+date to be June 30, 2024. This date matches the
+`end of extended life-cycle support <https://endoflife.software/operating-systems/linux/red-hat-enterprise-linux-rhel>`_
+from RedHat for RHEL 6. After this date, we build packages against
+CentoOS 7 by default for ``linux64``.
+
+
 2023-01-09: conda-forge Google Group is Now Read-only - Move to Discourse
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-We have made the conda-forge Google Group read-only. Please use the new 
-conda-forge `discourse forum <https://conda.discourse.group/c/pkg-building/conda-forge/25>`_, 
-our `Gitter room <https://gitter.im/conda-forge/conda-forge.github.io>`_, or it's `Matrix/Element 
+We have made the conda-forge Google Group read-only. Please use the new
+conda-forge `discourse forum <https://conda.discourse.group/c/pkg-building/conda-forge/25>`_,
+our `Gitter room <https://gitter.im/conda-forge/conda-forge.github.io>`_, or it's `Matrix/Element
 counterpart <https://app.element.io/#/room/#conda-forge-space:matrix.org>`_ instead.
 
 
 2023-01-08: ``conda-forge/staged-recipes`` Feedstock Creation Job Moved
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-We have moved the CI job that makes new feedstocks to our 
-`conda-forge/admin-requests <https://github.com/conda-forge/admin-requests>`_ 
+We have moved the CI job that makes new feedstocks to our
+`conda-forge/admin-requests <https://github.com/conda-forge/admin-requests>`_
 repo. The new location is reflected in the various links on repos and our status page.
 
 2022
--- a/src/user/announcements.rst
+++ b/src/user/announcements.rst
@@ -8,8 +8,8 @@ Our announcements are published to an RSS feed `here <https://conda-forge.org/do
 2023
 ----
 
-2023-07-12: EOL for CentOS 6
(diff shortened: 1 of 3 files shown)


src/user/announcements.rst
```diff
@@ -8,20 +8,31 @@ Our announcements are published to an RSS feed `here <https://conda-forge.org/do
...
 
+2023-07-12: End-of-life for CentOS 6
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+As you may be aware, we have delayed the deprecation of our CentOS 6 build
+system the ``linux64`` platform several times. We have now set a formal deprecation
+date to be June 30, 2024. This date matches the
+`end of extended life-cycle support <https://endoflife.software/operating-systems/linux/red-hat-enterprise-linux-rhel>`_
+from RedHat for RHEL 6. After this date, we build packages against
+CentoOS 7 by default for ``linux64``.
```
kkraus14, line 19:
```suggestion
CentOS 7 by default for ``linux64``.
```

jakirkham, line 19: This is the EOL for CentOS 6 & 7. So don't think we can move to CentOS 7 then

jakirkham, line 19:
```suggestion
CentosOS 7 by default for ``linux64``.
```

jakirkham, line 19:
```suggestion
CentosOS 7 by default for ``linux64``.
```

beckermr: I'm happy to change the wording here, but I think jumping past cos7 will be hard for the same reasons cos6 was. We will have a bunch of users with legacy systems and we don't formally need to jump ahead, so it makes more sense to stick with cos7.

beckermr: fixed!

beckermr: Also this isn't standard EOL. This date is for their extended lifecycle support maintenance. That goes until June 30, 2028 for cos7. See https://access.redhat.com/support/policy/updates/errata#Extended_Life_Cycle_Phase

The June 30, 2024 date is the end of extended lifecycle support for cos6. 

So going to cos7 in 2028 would be consistent.

jakirkham: That link is for RedHat. Not CentOS. The CentOS page confirms what the other link stated. Namely CentOS 7 is EOL 30 June 2024

https://wiki.centos.org/About/Product

<hr>

Appreciate that the situation is messy due to the RedHat changes. However we can't control that. All we can control is the safety of our own infrastructure, which gets significantly harder once CentOS 7 goes EOL. Are we planning to maintain CentOS 7 ourselves? What is our plan?

mattip:
With my hat of NumPy maintainer, this puts us in a difficult position. It means we should continue to support glibc-2.12 for another year (I seem to remember we already removed code to support it when CentOS6 went EOL at the end of 2020) and also support the manylinux2014 standard glibc-2.17 after manylinux2014 is "officially" EOL. We were looking forward to ripping out some work-arounds we have for buggy glibc-2.17 math routines.

chrisburr:
@mattip There is no need for numpy to continue to support glibc-2.12, we can override it on a per-feedstock basis and already do for many packages.

See https://conda-forge.org/docs/maintainer/knowledge_base.html#using-centos-7

People can continue to get updates of downstream packages and rely on us building with older numpy ABIs to still be able to use the newer downstream packages while using an older numpy.

chrisburr:
The only place we might have an annoying issue is for Python 3.12 as if there isn't a version of numpy that supports both 3.12 and CentOS 6 (but that's a purely conda-forge issue and I can think of workarounds).

jakirkham:
That said, if NumPy requires GLIBC 2.17+, are we gaining much by keeping GLIBC 2.12? Basically the PyData stack needs to move to GLIBC 2.17+

src/user/announcements.rst
jakirkham: Lol and my fix had a typo 🤣

Thanks Matt! 🙏

mattip:
I fear maintaining the patches to revert post glibc-2.17 changes to the [blocklisted math functions](https://github.com/numpy/numpy/blob/main/numpy/core/src/common/npy_config.h#L140) may be a large maintenance burden on conda-forge. Luckily, MSVC support mandates we keep some of these blocklisted functions around, but if Microsoft fixes those, we could drop them which would greatly clean up NumPy internals.

jakirkham:
Is there an open issue tracking NumPy's GLIBC update plans?

mattip:
Not really. There is [a numpy issue about npymath](https://github.com/numpy/numpy/issues/20880) with a wide-ranging discussion around many things, among them math routine support in the OS runtime. I did open an issue in the canonical guide to PEP 600 compliance mayeut/pep600_compliance#306 which so far has been my guide to manage manylinux (and thus glibc) deprecation

h-vetinari:
I think we shouldn't make an argument based on the extended EOL, because it's completely irrelevant to our infra (it's not like we'd get any updates, nor do we need them really). We just decided to support _compatibility_ with old Linux longer due to known users being affected.

I don't think we should establish an expectation that our support is related to the end of rhel's extended EOL. It was and remains based on our goodwill to support things as long as possible, but if a big problem arises we should be free to move ~more quickly~ less glacially.

src/user/announcements.rst
beckermr: Yes I know the link was for redhat. My point is simply that some people who use conda-forge may still be on cos7 even if it is past EOL.

beckermr:
I didn't mean to imply we should base our lifecycle on RedHat. I was trying to point out that even if CentOS7 is past EOL, folks still may be using it just like what happened with CentOS6. 

I 100% agree it is based on our maintenance capacity. 

We have to move to something when we deprecate CentOS6 and simply going to CentOS7 is by far the easiest thing. Individual feedstocks can move ahead as they please.

src/user/announcements.rst
beckermr: We are not making a promise of 4 years of cos 7 support with this announcement. We are simply saying that when we turn off Cos6, we will use cos 7. We can choose to deprecate cos7 a day later if we please.

PR #1988, opened 2023-07-25
added docs for swapfile switch

mikemhenry:
PR Checklist:

This PR is for adding docs if https://github.com/conda-forge/conda-smithy/pull/1751 is merged

### DIFF
--- a/src/maintainer/conda_forge_yml.rst
+++ b/src/maintainer/conda_forge_yml.rst
@@ -82,7 +82,7 @@ mapping for Azure-specific configuration options. For example:
       # Option for creating a swap file on the default Azure Pipelines
       # linux image before running the Docker container for building
       # for example, use 10G to create a 10G swap file
-      make_swap: 0
+      make_swap: 0G
       # limit the amount of CI jobs running concurrently at a given time
       # each OS will get its proportional share of the configured value
       max_parallel: 25
--- a/src/maintainer/conda_forge_yml.rst
+++ b/src/maintainer/conda_forge_yml.rst
@@ -79,6 +79,9 @@ mapping for Azure-specific configuration options. For example:
       # toggle for freeing up some extra space on the default Azure Pipelines
       # linux image before running the Docker container for building
       free_disk_space: False
+      # toggle for creating a 10G swap file on the default Azure Pipelines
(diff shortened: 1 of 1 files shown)


src/maintainer/conda_forge_yml.rst
```diff
@@ -79,6 +79,9 @@ mapping for Azure-specific configuration options. For example:
       # toggle for freeing up some extra space on the default Azure Pipelines
       # linux image before running the Docker container for building
       free_disk_space: False
+      # toggle for creating a 10G swap file on the default Azure Pipelines
+      # linux image before running the Docker container for building
+      make_10G_swap: False
```
jakirkham, line 84: Wonder if we should allow the size to be customizable. Something like this perhaps?

```suggestion
      make_swap_file: 10Gb
```

jakirkham:
Thanks Mike! 🙏

Had some thoughts on how the configuration could be presented

src/maintainer/conda_forge_yml.rst
mikemhenry: Yes this is a great idea, the default can be zero and if it is zero it just skips that whole block, but if greater than zero we make a swap file of that size

jakirkham: Yeah actually default of `0` makes more sense. Let's go with that

src/maintainer/conda_forge_yml.rst
```diff
@@ -79,9 +79,10 @@ mapping for Azure-specific configuration options. For example:
       # toggle for freeing up some extra space on the default Azure Pipelines
       # linux image before running the Docker container for building
       free_disk_space: False
-      # toggle for creating a 10G swap file on the default Azure Pipelines
+      # Option for creating a swap file on the default Azure Pipelines
       # linux image before running the Docker container for building
-      make_10G_swap: False
+      # for example, use 10G to create a 10G swap file
+      make_swap: 0
```
jakirkham, line 85:
```suggestion
      make_swap: 0G
```

jakirkham (approved):

jakirkham:
Thanks Mike! 🙏

PR #1989, opened 2023-07-27
Document some `outputs` use cases and pitfalls

jaimergp:
PR Checklist:

Comes from discussion at https://github.com/conda-forge/duckdb-feedstock/issues/6#issuecomment-1650705191

### DIFF
--- a/docs/maintainer/knowledge_base.md
+++ b/docs/maintainer/knowledge_base.md
@@ -1561,6 +1561,40 @@ add some information on r packages which make heavy use of `noarch: generic`
 
 :::
 
+## Multi-output recipes
+
+`conda-build` has the ability to create multiple package artifacts from a single recipe via the `outputs` section in `meta.yaml`. This is useful in several scenarios, including:
+
+- Distributing a project (which share the same source code) in separate artifacts. For example:
+  - A compiled C++ library and its Python bindings:
+    - [mamba-feedstock](https://github.com/conda-forge/mamba-feedstock/blob/main/recipe/meta.yaml)
+  - A runtime library and its headers:
+    - [cpp-opentelemetry-sdk](https://github.com/conda-forge/cpp-opentelemetry-sdk-feedstock/blob/main/recipe/meta.yaml)
+  - A dynamic library and a static version:
+    - [libarchive](https://github.com/conda-forge/libarchive-feedstock/blob/main/recipe/meta.yaml)
+- Distributing the same project with different sets of dependencies. For example:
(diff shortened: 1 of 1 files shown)


src/maintainer/knowledge_base.rst
```diff
@@ -1309,6 +1309,78 @@ Noarch generic
...
+  
+  - A runtime library and its headers:
+    
+    - `cpp-opentelemetry-sdk <https://github.com/conda-forge/cpp-opentelemetry-sdk-feedstock/blob/main/recipe/meta.yaml>`__
+
+  - A dynamic library and a static version:
+
+    - `libarchive <https://github.com/conda-forge/libarchive-feedstock/blob/main/recipe/meta.yaml>`__
+
+- Distributing the same project with different sets of dependencies. For example:
```
ocefpaf, line 1334: This is also a trick to mimic optional dependencies from Python-land. Maybe we should mentioned that here too?

ocefpaf (approved):
This is awesome @jaimergp! Thanks so much for this PR.

src/maintainer/knowledge_base.rst
jaimergp: Added! 🦸

dhirschfeld:
`typing-extensions` uses `{{ pin_subpackage(name, exact=True) }}`
`importlib-metadta` uses `{{ pin_subpackage('importlib-metadata', max_pin="x.x.x") }}`

but over in https://github.com/conda-forge/duckdb-feedstock/issues/6#issuecomment-1654025349 @isuruf said it should be `{{ version }} *_{{ build_num }}`

Should the preferred version be explicitly documented in the text?

dhirschfeld:
From testing (over in https://github.com/conda-forge/arrow_odbc-feedstock/pull/23) it appears it should be `={{ version }}=*_{{ build_num }}`

wshanks:
A couple suggestions based on my experience in https://github.com/conda-forge/qiskit-terra-feedstock/pull/28:

* If it should be possible to output a Python arch-specific package and a Python noarch package from a single feedstock, document how to do that. I guess in https://github.com/conda-forge/qiskit-terra-feedstock/pull/28 it is almost working other than the cross-compiling case. I wonder if anything weird happens with every arch building the same noarch package? Or maybe the noarch should be set to skip most of the runs?.
* Consider suggesting using `bot: check_solvable: true` as an alternative way to maintain tightly coupled packages with separate feedstocks without `outputs`, especially if the dependent package is simple and could be set to automerge?

xhochy (approved):

PR #1995, opened 2023-08-21
add announcement about miniconda & mambaforge unification

jaimergp:
PR Checklist:

xrefs:
- https://github.com/conda-forge/miniforge/issues/284
- https://github.com/conda-forge/miniforge/pull/277

### DIFF
--- a/src/user/announcements.rst
+++ b/src/user/announcements.rst
@@ -11,7 +11,8 @@ Our announcements are published to an RSS feed `here <https://conda-forge.org/do
 2023-08-21: Miniforge and Mambaforge installers are now identical
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-With the release of ``Miniforge3-22.3.1-0``, the Miniforge and Mambaforge installers are
+With the `release <https://github.com/conda-forge/miniforge/releases/tag/23.3.1-0>`__ 
+of ``Miniforge3-22.3.1-0``, the Miniforge and Mambaforge installers are
 now identical. The only difference between the two is the name of the installer and,
 subsequently, the default installation directory.
 
--- a/src/user/announcements.rst
+++ b/src/user/announcements.rst
@@ -8,6 +8,21 @@ Our announcements are published to an RSS feed `here <https://conda-forge.org/do
 2023
 ----
 
+2023-08-21: Miniforge and Mambaforge installers are now identical
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+With the release of ``Miniforge3-22.3.1-0``, the Miniforge and Mambaforge installers are
+now identical. The only difference between the two is the name of the installer and,
+subsequently, the default installation directory.
+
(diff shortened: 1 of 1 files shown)


ocefpaf:
I wonder if we should provide more info on why we decided to do this that may answer some questions.

Trying to put myself on the shoes of a user here I guess that they could ask:: Why the unification? Should I choose one or another going forward at the risk of one of them getting deprecated? If I used mini/mamba-forge am I wasting/saving space by getting more/less?

jaimergp:
Thanks @ocefpaf! Let me answer inline and then we can decide if the info is worth sharing in the announcement or maybe a FAQ section in [conda-forge.org/miniforge](https://conda-forge.org/miniforge/).

> Why the unification?

Miniforge is a conda-forge-based, Miniconda-like installer. It has historically tried to provide the same components. Miniconda started shipping conda-libmamba-solver in July, in preparation for the default solver switch in September. 

Bundling conda-libmamba-solver implies pulling the `libmambapy` dependency tree (libmamba, libsolv, libarchive, libcurl, etc), bringing it to parity with the Mambaforge contents (except for the tiny `mamba` Python modules).

Since the installer size difference is now moot, we decided to go ahead and merge both for simplicity and "less surprising" behavior. 


> Should I choose one or another going forward at the risk of one of them getting deprecated?

There are no concrete plans to deprecate any or the other. That said, if you were to ask _me_, I'd keep Miniforge and eventually deprecate Mambaforge. If we ever decide to do so (which might break too many unattended setups), it will be properly announced, with deprecation warnings and brown-outs or whatever we come up with.

> If I used mini/mamba-forge am I wasting/saving space by getting more/less?

If you were using Mambaforge already, this change won't have any perceivable effects. If you were stuck with Miniforge, this release opens up new fast solver possibilities, at the cost of some megabytes. It shouldn't be that terrible (~20MB in installer size for Linux x64).

ocefpaf:
Indeed that is the info some users may be after. I'm not sure if it is too verbose for that announcement or if it should be there. To be honest, I'm on the fence, but I know those things and the user reading the announcement may not.

jaimergp:
I added https://github.com/conda-forge/miniforge/pull/478 so we can link to the FAQ section of the Miniforge repo.

src/user/announcements.rst
```diff
@@ -8,6 +8,22 @@ Our announcements are published to an RSS feed `here <https://conda-forge.org/do
...
+now identical. The only difference between the two is the name of the installer and,
+subsequently, the default installation directory.
+
+From now on, both installers will provide:
+
+- ``conda`` (23.3.1 in this release)
+- ``conda-libmamba-solver`` (23.3.0 in this release)
+- ``mamba`` (1.4.2 in this release)
+- ... plus their dependencies
+
```
hmaarrfk, line 25: I would only add a note that we are discouraging the use of Mambaforge

src/user/announcements.rst
```diff
@@ -8,6 +8,22 @@ Our announcements are published to an RSS feed `here <https://conda-forge.org/do
...
+
```
hmaarrfk, line 26:
```suggestion

We have decided to keep the download links
of Mambaforge active to help those that have automated workflows.

The new usage of Mambaforge is thus discouraged. Bug
reports specific to Mambaforge will be closed as won't fix.
```

PR #2068, opened 2024-01-31
Move meeting minutes to community/meeting-minutes

goanpeca:
Related to https://github.com/conda-forge/conda-forge.github.io/issues/1971

To be merged after https://github.com/conda-forge/conda-forge.github.io/pull/2066

PR Checklist:

- [x] note any issues closed by this PR with [closing keywords](https://help.github.com/articles/closing-issues-using-keywords)
- [x] put any other relevant information below

## `/community/meeting-minutes/`

Screenshot 2024-01-31 at 7 07 37 PM

## `/community/meeting-minutes/2024-01-24/`

Screenshot 2024-01-31 at 7 10 44 PM

PR #2066, opened 2024-01-30
Move org files from sphinx to community/orgs

goanpeca:
Related to https://github.com/conda-forge/conda-forge.github.io/issues/1971

- Move docs/orga to community/organization

PR Checklist:

- [x] note any issues closed by this PR with [closing keywords](https://help.github.com/articles/closing-issues-using-keywords)
- [x] put any other relevant information below

## Work made

- Content from `/docs/orga/` has been migrated to `/community/`,  except for `funding`, `guidelines` and `meeting minutes`. These will be organized in a subsequent PR
- RST files were converted to GFM (github flavored markdon) using pandoc
- Order of pages was adjusted with frontmatter
- Generation script for cfp index was updated
- Generation scripts for members was created
- Footer links and redirects were updated as well as necessary docusaurus config

### `/community/`

Screenshot 2024-01-30 at 6 58 27 PM

### `/community/governance/`

Screenshot 2024-01-30 at 6 58 32 PM

### `/community/join/`

Screenshot 2024-01-30 at 6 58 38 PM

### `/community/subteams/`

Screenshot 2024-01-30 at 6 58 46 PM

### `/community/cfep/`

Screenshot 2024-01-30 at 6 58 53 PM

### `/community/get-in-touch/`

Screenshot 2024-01-30 at 6 59 02 PM

### `/community/contracting/`

Screenshot 2024-01-30 at 6 59 10 PM

Issue #1971, opened 2023-06-30
Redesigning conda-forge.org


jaimergp:
Hello conda-forge!

We now have a new website prototype available at https://cf-infra-docs.netlify.app/. This mostly showcases the work @asmitbm did during his GSoC internship last summer, but you can also see how some documentation would look like there, or the blog.

This prototype is based on the Docusaurus framework, which uses a React, Node and Markdown. 

In parallel, @afshin has been working on a new status page in [this PR](https://github.com/Quansight-Labs/cf-infra-docs/pull/12). [A preview is also available](https://deploy-preview-12--cf-infra-docs.netlify.app/status/).

I've also done some work studying potential migration strategies for our documentation. I don't want to lose the valuable git history by converting all the RST over to MD, so instead I suggest the following strategy:

1. Keep Sphinx around for now, but have it render to Markdown instead of HTML (via `sphinx-markdown-builder`).
2. Have Docusaurus take those Markdown documents and render the HTML with them.
3. Whenever needed, slowly migrate RST content to native MD in Docusaurus. This might never happen though, if we feel that Sphinx's linking abilities are too important. We can also end up in a hybrid situation, with some narrative documentation being rendered natively in Docusaurus, and some more reference-y docs done in Sphinx.

I have a demo of how this strategy works in https://github.com/Quansight-Labs/cf-infra-docs/pull/14. [Plus the corresponding preview](https://deploy-preview-14--cf-infra-docs.netlify.app/docs/).

---

### Migration steps

This is my proposed timeline:

- [x] 1. Implement Docusaurus in this repository, starting with the frontpage _only_. Ensure there are no broken links - https://github.com/conda-forge/conda-forge.github.io/pull/2050
- [x] 2. Migrate the blog ([the preview site already has it](https://cf-infra-docs.netlify.app/blog/)), while keeping git credits around.
- [x] 3. Migrate `feedstock-outputs`
  - [x] Frontend: https://github.com/conda-forge/conda-forge.github.io/pull/2064
  - [X] Backend should produce JSON: https://github.com/conda-forge/feedstock-outputs/pull/58
- [x] 4. Migrate the Sphinx documentation as discussed above.
  - [x] Migrate announcements to a new blog-like instance - https://github.com/conda-forge/conda-forge.github.io/pull/2057
  - [X] Migrate most of the Orga content to a new community docs-like instance: https://github.com/conda-forge/conda-forge.github.io/pull/2093
[...]

PR #2077, opened 2024-02-13
Render Sphinx `/docs` in docusaurus

jaimergp:
PR Checklist:

---

- Part of #1971.
- #2066 and #2068 should probably go in before this one.
- Based on https://github.com/Quansight-Labs/cf-infra-docs/pull/14. 
- How it works:
  - This uses a fork of `sphinx-markdown-builder` that adds better Docusaurus support (e.g. admonitions) to render the Sphinx docs to markdown documents.
  - The rendered Markdown has resolved all the references and Sphinx directives to regular Markdown Docusaurus can understand.
  - These MD documents are moved to `/docs` by a special script that also handles some minor postprocessing (adding `sidebar_position`, rewriting some URLs, etc).
  - From there, Docusaurus will take `/docs` as if they were written by hand and publish the site.
- The default URL structure is a bit different in docusaurus. Directory names like `/path/thing/` are preferred to file names like `/path/thing.html`. Our Sphinx site also had strange `00_intro.html` instead of `index.html`. To that effect, we have added redirects from `/path/to/thing.html -> /path/to/thing/` and `/path/00_intro.html -> /path/`.
- Since we have now all the site rendered under Docusaurus, we can add some plugins:
  - Algolia search for super awesome search results. These should become available once we merge. It should look like the search bar in conda.org. If we don't want Algolia, we can use `lunr` (local search, JS plugin, not as powerful).
  - GoatCounter for OSS-friendly stats, no cookies 

Notes for reviewers:

- We have PR previews now! Check how the new docs look like at https://deploy-preview-2077--conda-forge-previews.netlify.app/docs/.
- There might be minor formatting issues so hopefully we catch those during the review. Screenshots or links in the preview would be super helpful!
- Yes, the `sphinx-markdown-builder` fork situation is a bit fragile. We can try to upstream things but: (1) some things are Docusaurus specific and might need options; (2) this is supposed to be temporary to some degree, because ideally we would end up writing everything in native Docusaurus 🤞. If (2) doesn't happen, then yes, upstreaming is even more important.


.ci_scripts/sphinx_markdown_to_docusaurus.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+import sys
+from pathlib import Path
+
+from ruamel.yaml import YAML
+
+yaml = YAML(typ="safe")
+yaml.default_flow_style = False
+
+SIDEBAR_ORDERING = [
```
beckermr, line 12: If we change the website's high-level structure, does that mean this list will need to change too? If yes, we should add a comment and maybe have some sort of CI check.

.ci_scripts/update_docs
```diff
@@ -2,35 +2,27 @@
...
 
-python .ci_scripts/generate_cfep_index.py
-
-pushd sphinx/src
+python "$BASE_DIR/.ci_scripts/generate_cfep_index.py"
 
+pushd "$BASE_DIR/sphinx/src"
+make clean
 # -W --keep-going: list all warnings but fail build in case there are any
-# -n: check validity of all links
```
beckermr, line 17: I guess we cannot run link check anymore? This has caught many bugs for us.

beckermr (commented):
LGTM from me but I left a few comments.

.ci_scripts/update_docs
jaimergp: Docusaurus runs it later (and much faster), so no need to do it with Sphinx. Let me push a broken link on purpose to show it in action.

beckermr: No need to test it. I didn't know! Thank you!

jaimergp: Well, turns out it only check internal links, no external stuff. Adding a new step for now.

.ci_scripts/sphinx_markdown_to_docusaurus.py
jaimergp: Good catch! Added a check to raise if the document is not found in this list. Hopefully, if we start changing `sphinx/` stuff, is to bring it over to Docusaurus, so "this should not happen" but better be safe than sorry.

beckermr:
Well I tried to fix the core and emeritus member lists @jaimergp but I think I am missing something important about how sphinx is working now or I got the relative path wrong.

sphinx/src/conf.py
```diff
@@ -355,10 +358,10 @@ def get_formated_names(path_file):
     current_file = os.path.dirname(__file__)
     context = app.config.html_context
     context["core_members"] = get_formated_names(
-        os.path.join(current_file, "..", "..", "src", "core.csv")
+        os.path.join(current_file, "..", "..", "..", "src", "core.csv")
```
beckermr, line 361:
```suggestion
        os.path.join(current_file, "..", "..", "..", "..", "src", "core.csv")
```

sphinx/src/conf.py
```diff
@@ -355,10 +358,10 @@ def get_formated_names(path_file):
...
     )
     context["emeritus_members"] = get_formated_names(
-        os.path.join(current_file, "..", "..", "src", "emeritus.csv")
+        os.path.join(current_file, "..", "..", "..", "src", "emeritus.csv")
```
beckermr, line 364:
```suggestion
        os.path.join(current_file, "..", "..", "..", "..", "src", "emeritus.csv")
```

jaimergp:
Let me take a look, you mean the governance.rst substitution, right?

jaimergp:
Ok, the early return branch made us skip that part because the rendered is not HTML anymore, but Markdown. I adjusted a couple things and also added the CFEP substitution there because it's cleaner (otherwise you might risk accidentally committing the rendered version, like it happened two days ago :D).

jaimergp:
Fixed! See:

* https://deploy-preview-2077--conda-forge-previews.netlify.app/docs/orga/governance/#current-members-of-core
* https://deploy-preview-2077--conda-forge-previews.netlify.app/docs/orga/cfep-index/

beckermr:
Awesome. This last thing was the only comment on slack. I'd say lgtm and let's merge this. Any other issues can be fixed later.

jaimergp:
Yes, I can take a look on the search bar too. Hopefully it starts working once we merge. I don't see any difference with conda.org's setup.

PR #2084, opened 2024-02-23
Freeze Sphinx content to Markdown

jaimergp:
PR Checklist:

- Part of #1971 
- In the [last conda-forge core call meeting](https://github.com/conda-forge/conda-forge.github.io/pull/2083), we discussed the pros and cons of removing the render-to-markdown pipeline altogether because the maintenance complexity was not worth in the long run and Sphinx was not providing any features we were really missing in Docusaurus. In this PR I have rendered `/docs` permanently and reworked some `00_intro.html` pages with Docusaurus-equivalent features (card lists, mostly).


.readthedocs.yaml
```diff
@@ -1,7 +0,0 @@
-version: 2
```
beckermr, line 1: We had this to ensure the RTD pages redirect properly IIRC. Does that still work?

The RTD link is: https://conda-forge.readthedocs.io/en/latest/

beckermr (changes requested):
It appears that some of the static content images have been moved? This may break the status page again.

beckermr:
We should either keep the links the same or go back and fix the status page later.

.readthedocs.yaml
jaimergp: Ah, I didn't know we had that! Sorry and thanks!

jaimergp:
> It appears that some of the static content images have been moved? This may break the status page again.

When we fixed it last time we [pinned the commit hash](https://github.com/regro/cf-scripts/pull/2125/files#diff-94f65b30b318493facd1742635a5dd2fe36e57af38f9b2b75e78dd1424e8c9daR381), so it should be ok.

beckermr (approved):

goanpeca (approved):

zklaus (approved):
Looks good to me. Thanks for doing this!

PR #2087, opened 2024-02-26
Rework glossary

jaimergp:
PR Checklist:

- Part of #1971
- Closes #1363 
- Closes #1395


beckermr (commented):
Is there a way to automatically add glossary links instead of people having to add them by hand?

jaimergp:
> Is there a way to automatically add glossary links instead of people having to add them by hand?

Like in Sphinx with things like `:term:something`? Unfortunately no. It has to be done by "hand". There is this plugin (https://github.com/grnet/docusaurus-terminology) but (1) I haven't tried it, and (2) honestly I don't know if it's worth it given the extra setup (each term must be its own separate page). It's not like we have used the glossary too much either so I refrained from getting fancy here. If we see that its usage gets in the way, we can iterate and craft something nice. WDYT?

PR #2089, opened 2024-02-26
Use explicit sidebar configuration

jaimergp:
PR Checklist:

- Follow up to #2084 
- Part of #1971 

Use a explicit configuration for the docs/ sidebar so we don't have to maintain a set of relative `sidebar_position`. This is more similar to what we had in Sphinx. Also added a Python script to ensure we are not missing any.


.ci_scripts/check_sidebars.py
```diff
@@ -0,0 +1,51 @@
...
+                continue
+            path = str(f.relative_to(DOCS_DIR))[:-len(ext)]
+            if path.startswith("orga/minutes/"):
+                continue
+            if path.startswith("orga/funding/"):
+                continue
+            files.append(path)
+    return files
+
+def get_sidebar_files():
```
beckermr, line 22: Why isn't this function recursive since subitems also have link and id bits.

beckermr (changes requested):

.ci_scripts/check_sidebars.py
jaimergp: Good point! I initially went for the bare minimum (we don't have nested categories yet), but this will make it more future proof. I also added support for some syntax we are not using yet to prevent errors down the line.

beckermr (approved):

PR #2090, opened 2024-02-26
Status Dashboard

afshin:
This is an updated status/dashboard page for the conda-forge website.

This PR covers the functionality every core component of the current conda-forge status site.

Please [TRY IT OUT](https://deploy-preview-2090--conda-forge-previews.netlify.app/status/), feedback is welcome 🙏

cf. https://github.com/conda-forge/conda-forge.github.io/issues/1971


afshin:
The CI failure seems to come from trying to access a dynamic URL, but I am not certain. You can see in the deployment that the actual site logic is functional, but there might be something I am missing.

traversaro:
> Please [TRY IT OUT](https://deploy-preview-2090--conda-forge-previews.netlify.app/status/), feedback is welcome 🙏

Thanks a lot! As a "normal" contributor, I was wondering if in this new version there is a way to see the error messages for "non-solvable" packages in migration, see for example https://conda-forge.org/status/#libabseil20240116_libgrpc161_libprotobuf4252 .

src/pages/status/index.jsx
```diff
@@ -0,0 +1,16 @@
...
+import Layout from "@theme/Layout";
+import StatusDashboard from "@site/src/components/StatusDashboard";
+
+export default function Home() {
+  const { siteConfig } = useDocusaurusContext();
+  return (
+    <Layout
+      title={siteConfig.title}
+      description="Status dashboard for conda-forge"
+    >
```
jaimergp, line 12:
```suggestion
import Layout from "@theme/Layout";
import StatusDashboard from "@site/src/components/StatusDashboard";

export default function Home() {
  return (
    <Layout
      title="Status dashboard"
      description="Status dashboard for conda-forge"
    >
```

The title gets added automatically to the siteConfig one.

src/components/StatusDashboard/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+    // When all components finish loading, scroll if necessary.
+    if (jumped || loaded !== total) return;
+    setState((prev) => ({ ...prev, jumped: true }));
+    const id = hash.length > 1 ? hash.substring(1) : "";
+    if (id) document.getElementById(id)?.scrollIntoView();
+  });
+  const onLoad = () =>
+    setState((prev) => ({ ...prev, loaded: prev.loaded + 1 }));
+  return (
+    <main className={["container", styles.status_dashboard].join(" ")}>
```
jaimergp, line 49: I think the container is not working as intended now. It does not center on my big display. See https://github.com/conda-forge/conda-forge.github.io/blob/6cbd394342097effd33c0c7ebee9ff66748964a9/src/components/Packages/index.jsx#L282-L284 for an example on what worked for me in other pages.

src/components/StatusDashboard/current_migrations.jsx
```diff
@@ -0,0 +1,268 @@
...
+            onClick={() => resort("name")}
+            className={sort.by === "name" ? styles[sort.order] : undefined}
+          >
+            Name
+          </th>
+          <th
+            onClick={() => resort("status")}
+            className={sort.by === "status" ? styles[sort.order] : undefined}
+          >
+            Status
```
jaimergp, line 120: Can we call this "Completion rate"? Status might be ambiguous,

src/components/StatusDashboard/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+            </div>
+          </div>
+          <div className="row row--no-gutters">
+            <div className="col col--12">
+              <TravisCIUsage onLoad={onLoad} />
+            </div>
+          </div>
+          <div className="row row--no-gutters">
+            <div className="col col--6">
+              <Incidents onLoad={onLoad} />
```
jaimergp, line 85: One feature of the current status page is that when there's an ongoing incident, this panel gets promoted to the very top of the dashboard for visibility. I think this is not implemented here, right?

src/pages/status/index.jsx
jaimergp, line 1: I don't know how common this is in the JS world, but I'm missing the equivalent of module docstrings in almost all files. Having a little bit of technical documentation to see how to navigate the dashboard code would be super helpful.

src/components/MigrationDetails/styles.module.css
jaimergp, line 1: I think you were trying to reuse the same colors provided by the progress bar, but I think those were not deliberately chosen, but inherited from matplotlib default schema. They are not passing accessibility tests so I'm wondering if we can rely on the system colors we have at https://conda-forge.org/style-guide/#additional-colors. We might need to add more, though.

src/components/StatusDashboard/current_migrations.jsx
```diff
@@ -0,0 +1,268 @@
...
+        {rows.map((row) => {
+          const { progress } = row;
+          const href = `/status/migration/${row.name}`;
+          return (
+            <tr key={row.name}>
+              <td>
+                <a className="badge badge--secondary"
+                  href={href}
+                  style={{ minWidth: "100%" }}
+                  onClick={event => {
```
jaimergp, line 164: There's a chance this is preventing the "active" state from the `a` tags from working. Or if it's not this, something else is preventing that.

src/components/StatusDashboard/current_migrations.jsx
```diff
@@ -0,0 +1,268 @@
...
+        </tr>
+      </thead>
+      <tbody className={collapsed ? styles.collapsed  : undefined}>
+        {rows.map((row) => {
+          const { progress } = row;
+          const href = `/status/migration/${row.name}`;
+          return (
+            <tr key={row.name}>
+              <td>
+                <a className="badge badge--secondary"
```
jaimergp, line 161: Not sure if we want badges for these links. I liked it better when they were plain links.

jaimergp (commented):
Added some comments with, mostly, tiny things. The biggest item is that there's no documentation in the code at all. While it's well written and it can be easily navigated in the IDE, I think it would be beneficial to have a big picture description on how the dashboard is built and under what principles. If possible, each module should have a top-level docstring (or whatever the equivalent is in JS) with the purpose and scope of the module.

docusaurus.config.js
```diff
@@ -408,8 +422,8 @@ const config = {
         copyright: copyright,
       },
       prism: {
-        theme: prism.themes.github,
-        darkTheme: prism.themes.dracula,
+        theme: themes.github,
+        darkTheme: themes.dracula,
```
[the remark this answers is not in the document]
afshin: This failed to build for me this way, possibly because the version was very old. Was it actually building for you?

afshin:
> > Please [TRY IT OUT](https://deploy-preview-2090--conda-forge-previews.netlify.app/status/), feedback is welcome 🙏
> 
> Thanks a lot! As a "normal" contributor, I was wondering if in this new version there is a way to see the error messages for "non-solvable" packages in migration, see for example https://conda-forge.org/status/#libabseil20240116_libgrpc161_libprotobuf4252 .

You're right, this is a piece of functionality that is missing in the new version. I'll give it some thought and try to incorporate it. (cc: @GabrielaVives)

afshin:
> I can't find that functionality either. The URL for that migration would be https://deploy-preview-2090--conda-forge-previews.netlify.app/status/migration/libabseil20240116_libgrpc161_libprotobuf4252. If I filter for unsolvable only, I just get:
> 
> image
> 
> Weirdly, the package names in the first column are not clickable (they take you nowhere), despite having a hover state. Don't know if bug or intended, but the hover is confusing.

I see what's happened. Almost all the left-column items have a PR link. Some do not, but they're being automatically linked to empty string. I will put in a check to only link the ones that actually have a URL for the PR.

beckermr (changes requested):
I cannot find the solver errors we post to the status page.

src/components/MigrationDetails/index.jsx
```diff
@@ -0,0 +1,237 @@
...
+function Table({ details }) {
+  const [filters, setState] = useState(ORDERED
+      .reduce((filters, [key]) => ({ ...filters, [key]: 0 })));
+  const feedstock = details._feedstock_status;
+  const rows = ORDERED.reduce((rows, [status]) => (
+    filters[status] ? rows :
+      rows.concat((details[status]).map(name => ([name, status])))
+  ), []);
+  return (
+    <>
```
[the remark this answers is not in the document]
afshin: `<></>` is shorthand for `<React.Fragment></React.Fragment>` and it is for when you want to return more than one root node, but each component must only have one root node. So you can use a fragment in order to not create a new `<div>` or some other DOM layer. They do not render, but they allow you to return, e.g., **two** divs (like I do with the anchor for the table of contents immediately above the actual component card).

beckermr (changes requested):
We also need to integrate a large banner at the top with the status repo issues tracker.

mfansler:
> image

Really neat! I have some feedback on the migrations views:

- I found the interaction for filtering by status counterintuitive. I expected clicking "In PR" to filter to only those with that status, rather than filtering out those (since everything is shown by default).

- The order of the filter buttons and the sets in the bar graph neglect the semantics. The order should be:

    ```
    Done > In PR > Awaiting PR > Awaiting parents > Not solvable > Bot error
    ```

    Note that this order is reflected in the viridis-like rainbow colors.

- Something that doesn't work on the current dashboard and keeps the same behavior in the update is the hyperlinking of nodes in the graph view. If one opens the graph image in a new tab/window, the hyperlinks become active and let one jump to the PR. This is perhaps the most valuable functionality in this status board for me in terms of moving migrations forward. It would be nice if the hyperlinks worked in the page.

src/constants.js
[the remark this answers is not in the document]
afshin: I am open to moving these things into some other namespace. My rationale for this file was:

- I found it difficult to find every end point the original site was hitting and as I started collecting them, I saw they were coming from multiple sources and for the sake of the next upgrade, I thought having a sort of API manifest was a good idea.
- Since we hit so many different services on different domains, I think this manifest gives us a good target for consolidating end points if that ever becomes possible.

Putting the charts configuration there seemed okay. Perhaps it should be somewhere else, though?

About URLs:
- One of the end points was failing because the site had been redirected permanently (`quay.io` => `redhat.com`), I am not sure whether it's easier to fix such things now that the API end point URLs are in one place, but it seems like maybe?
- The repos and bots URLs were actually parsing HTML versions of APIs that had JSON end points available.

src/components/StatusDashboard/current_migrations.jsx
afshin: I'll have to investigate. This is what prevents the `404` flash.

src/constants.js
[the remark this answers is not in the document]
afshin: The other reason I thought having all the URLs in the same place would be good is because I was considering different strategies for keeping the data on the page up-to-date. For example, there could be a refresh button in each header, or there could be a poll, or some other mechanism. It might help to have a changing timestamp query string added to all URLs, etc.

SylvainCorlay:
> Something that doesn't work on the current dashboard and keeps the same behavior in the update is the hyperlinking of nodes in the graph view. If one opens the graph image in a new tab/window, the hyperlinks become active and let one jump to the PR. This is perhaps the most valuable functionality in this status board for me in terms of moving migrations forward. It would be nice if the hyperlinks worked in the page.

Wow I did not know about this feature!

jaimergp:
> The CI failure seems to come from trying to access a dynamic URL, but I am not certain. You can see in the deployment that the actual site logic is functional, but there might be something I am missing.

The error is:

```
[404] https://conda-forge.org/status/migration/ | Failed: Network error: Not Found
```

Which is expected because this PR hasn't been merged yet so it's actually checking the old single page `/status`. So we can ignore for now, or remove this line: https://github.com/conda-forge/conda-forge.github.io/blob/6cbd394342097effd33c0c7ebee9ff66748964a9/.github/workflows/deploy.yml#L58

src/components/StatusDashboard/styles.module.css
```diff
@@ -0,0 +1,303 @@
...
+}
+
+.status_dashboard {
+  margin: 0;
+  max-width: 100%;
+}
+
+.status_dashboard_toc {
+  border: var(--ifm-table-border-width) solid #ccc;
+  border-radius: 5px;
```
jaimergp, line 138: Can we remove this border to match the style in `/docs`?

src/components/StatusDashboard/repos_and_bots.jsx
```diff
@@ -0,0 +1,130 @@
...
+        console.warn(`error loading web services status`, error);
+      }
+    })();
+  }, []);
+  return (
+    <tr>
+      <td>
+        <a href={urls.repos.services.link}
+          style={{ display: "inline-block", minWidth: "100%" }}>
+          admin web services
```
jaimergp, line 117: This should be either `webservices` or `conda-forge-webservices`, to make it match the repo name.

src/components/StatusDashboard/current_migrations.jsx
jaimergp: Got it. Let's leave it as is provided we can fix the hover/active states.

src/components/MigrationDetails/index.jsx
```diff
@@ -0,0 +1,237 @@
...
+  );
+}
+
+function Graph(props) {
+  const [error, setState] = useState("");
+  const url = urls.migrations.graph.replace("<NAME>", props.children);
+  const onError = (error) => setState(error);
+  return (
+    <div>
+      {error ? `Graph is unavailable.` : <img onError={onError} src={url} />}
```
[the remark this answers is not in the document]
SylvainCorlay: This could be in a separate PR.

jaimergp:
I've also noticed that the migration tables can only be sorted together; e.g. if I sort "long-running migrations" by name, "Regular migrations" and "Closed migrations" also get sorted in the same way:


https://github.com/conda-forge/conda-forge.github.io/assets/2559438/511d983b-e051-4c58-a095-fd47440ebeea

afshin:
These are not separate tables. They are one table with different
subheadings. This sorting behavior is by design. Does it harm usability to
sort the entire table by the column you’re interested in?

jaimergp:
Not necessarily. I was expecting individual behavior given the sort buttons in each subtable, but if it makes everything more complicated, for me it's not personally worth "fixing" it.

GabrielaVives:
> Not necessarily. I was expecting individual behavior given the sort buttons in each subtable, but if it makes everything more complicated, for me it's not personally worth "fixing" it.

I think that what is counter-intuitive here, is the repetitive column headers that give the impression that there are 3 independent tables. 
Maybe by having only one line with the column titles at the top, and the 3 dividers below would make it intuitive? 
The collapsable feature of each subheader would still allow for an easier (or quicker) way for users to access the column titles to filter.

GabrielaVives:
> * I found the interaction for filtering by status counterintuitive. I expected clicking "In PR" to filter to only those with that status, rather than filtering out those (since everything is shown by default).
> *

I agree, even though I thought it was an intuitive way in the first place, now that I am using it here I also don't expect it to behave the way it is. 
I think there also is some UI design work to be done to improve it, so I am looking in to it.

src/components/MigrationDetails/index.jsx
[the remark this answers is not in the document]
jaimergp: That works too. It's true that for the sake of scoping, this PR should only try to replicate the current functionality found in the old `/status`. Improvements can be done later.

beckermr:
This test issue flushed out some bugs: https://github.com/conda-forge/status/issues/167

It says the incident is resolved when it is not.
The date doesn't work.
The banner color is wrong for the issue type.
We should have something big that bubbles up to the top of the page.

jaimergp:
Screenshot for reference: 
image

beckermr:
Also, I think we should use the following colors for the different statuses:

operational: #0DE877
degraded performance: #FFA500
maintenance/investigating: #1192FC
major outage: #FF4D4D

jaimergp:
We have a semantic palette at https://conda-forge.org/style-guide/#additional-colors. If possible, we should stick to those, since they have been optimized for both dark and light mode, and they meet accessibility criteria. We might need one more shade, though.

beckermr:
Ahhh yes that's great! 

My real goal is that degraded performance should be "organge" not "red". 

So this works

- investigating/maintenance: #1976d2
- operational: #388e3c
- degraded performance: #ef6c00
- major outage: #d84315

beckermr:
Also, do we want to maintain the issue tracker on the status repo or maybe we should make status issues here in this repo?

afshin:
> I can't find that functionality either. The URL for that migration would be https://deploy-preview-2090--conda-forge-previews.netlify.app/status/migration/libabseil20240116_libgrpc161_libprotobuf4252. If I filter for unsolvable only, I just get:
> 
> image
> 
> Weirdly, the package names in the first column are not clickable (they take you nowhere), despite having a hover state. Don't know if bug or intended, but the hover is confusing.

Now it detects when there is no link available and it allows expanding for more information.

Screenshot 2024-03-01 at 12 23 19

beckermr:
I noticed the current status indicators for Travis etc don't indicate if maintenance is happening. 

Also the rendering on mobile is not great yet.

afshin:
> Also, do we want to maintain the issue tracker on the status repo or maybe we should make status issues here in this repo?

Perhaps changing the `status` repo to being *just* a repo for status issues and a `README` that describes what the repo is for would be a good continuity and clear landing spot? I like the shape of the URL `github.com/conda-forge/status` and the fact that it's already the de facto source for this is a good reason to continue.

Of course, this is just an opinion and I'm happy to go with whatever you think.

afshin:
Oof, but this seems blocked in the deployed version. This must be some sort of CORS related problem.

docs/sidebar.json
jaimergp, line 1: Revert this. To work around the sidebar checker errors (I have a PR in the making improving the DX), use `.ci_script/update_docs` at least once locally and then it should work.

afshin:
> I've noticed something weird with the navigation now. If you access the migration details on any row, and try to go back, you don't go back to the migration table; you skip it and go to whatever was there before. See video:
> Screen.Recording.2024-02-29.at.00.54.13.mov

This is fixed now. It was because in order to stay inside the React application, I needed to use `<Redirect>` but without specifying that you want it to `replace` your browser history when you instantiate it, it will just do redirect without adding it to the history queue. We have two redirects in this dashboard, one of them *should not* add to history (if you go to the migration details view but there is no migration, it redirects you back) and the other *should* replace the history (when you redirect *to* the migration details page and you want history to work):
```ts
// Use `replace` (equals `true`) prop for browser history.
if (redirect) return (<Redirect to={redirect} replace />);
```

beckermr:
This is getting really close @afshin! My only remaining request is that if there is an ongoing incident, we need to float a larger banner to the top of the status page with the title, color, etc.

mfansler:
> > * Something that doesn't work on the current dashboard and keeps the same behavior in the update is the hyperlinking of nodes in the graph view. If one opens the graph image in a new tab/window, the hyperlinks become active and let one jump to the PR. This is perhaps the most valuable functionality in this status board for me in terms of moving migrations forward. It would be nice if the hyperlinks worked in the page.
> 
> 
> 
> – @mfansler
> 
> 
> 
> I was rendering the graph as an `img`, so I switched to a library called `react-svg` which allows me to load the SVG as a React component – which includes its embedded hyperlinks, unlike a basic `img` tag:
> 
> 
> 
> Screenshot 2024-03-01 at 15 49 49
> 
> 

Fantastic! Thanks for addressing this! ❤️

beckermr:
Also I realized since this site is static, we'll need to push a rebuild for changes on the status repo. I can add a workflow to the status to push an empty commit here or bump a deploy workflow.

afshin:
I am unsure what kind of changes you’re considering. What sort of changes
would require a rebuild? Do you mean changes to the docs in this repo or do
you mean some other sort of changes? The site is static insofar as it’s not
server generated content, but upon load it grabs everything using live
APIs, so every refresh potentially contains new content.

jaimergp:
> My comment was exactly backwards, I have updated it and the code and hopefully this is correct now.

Ok, it works now! 🚀

beckermr:
Ahhhh @afshin. The changes I meant were to the issues, their content, their status, and labels on the conda-forge/status repo. I guess you are saying the user-side browser reloads that info via the GH API each time it loads the docs? If so, then we don't what I am saying.

SylvainCorlay:
This is looking good. Further improvement could be made in subsequent PRs.

jaimergp:
The 404s should go away if you remove this line: 

https://github.com/conda-forge/conda-forge.github.io/blob/ac4b5a43d11a30ff9da2ba802121ee8fb995e0b5/.github/workflows/deploy.yml#L60

beckermr (changes requested):
Please do not merge. The most important piece of information on the status page is the status of conda forge. Right now people have to scroll all the way to the bottom. This needs to be fixed.

package.json
```diff
@@ -18,12 +18,19 @@
...
+    "install": "^0.13.0",
+    "moment": "^2.29.4",
+    "octokit": "^3.1.2",
     "prism-react-renderer": "^2.3.1",
     "react": "^18.2.0",
-    "react-dom": "^18.2.0"
+    "react-chartjs-2": "^5.2.0",
+    "react-dom": "^18.2.0",
+    "react-markdown": "^9.0.1",
+    "remark-gfm": "^4.0.0"
```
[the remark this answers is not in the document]
jaimergp: Coming back to this, it [looks like what folks do](https://stackoverflow.com/questions/14221579/how-do-i-add-comments-to-package-json-for-npm-install) is something along the lines of:

```json
"packages": {
...
},
"// packages": {
  "//": "There's nothing special about this dict; it's just a collection of comments",
  "install": "This package is used for ...",
  ...
}
```

jakirkham:
Could I ask that we review the status page at the conda-forge meeting this week?

afshin:
> Please do not merge. The most important piece of information on the status page is the status of conda forge. Right now people have to scroll all the way to the bottom. This needs to be fixed. 

I will definitely have this done for the condo-forge meeting. I think this is almost ready for merging, but the incident handling needs to be at least at feature parity first.

beckermr:
Awesome! Thank you!

jakirkham:
When we looked at this page in the past, wrote up some notes here

Let's double check that those are addressed (and if not make the needed changes)

https://github.com/conda-forge/conda-forge.github.io/issues/1971#issuecomment-1865810748

beckermr:
That looks great. Is it possible to only render the open incidents at the top and leave the history at the bottom?

jaimergp:
Awesome!

I think there are couple of small nits, though:

image


The "ongoing" badge colors seem different but have the same "ongoing" label. I think that's different labels in the repo (major outage, degraded performance) 🤔 I think it would be better to just use the label text when the issue is open, and then fall back to "Resolved" once closed.

Also,  the timestamps are not correct, I think? They seem to point at the last modification time, not creation time, but then issues are sorted by number... that's a bit confusing. And speaking of timestamps: what's the timezone? local? utc? Would be nice to reflect that.

jaimergp:
Oops, Matt beat me to it.

beckermr:
nice find @jaimergp on the time stamps!

also jinx!

GabrielaVives:
I have been through the multiple comments giving feedback about the filters and colors in the migration details section.
There are several issues I would like to address:

- I think that the filtering buttons should have neutral colors rather than success or warning colors because their primary function is to categorize or sort data, not to convey a sense of achievement or caution. It will help promote clarity in this page where a lot of information is displayed, consistency, accessibility, and a focus on the primary function of the buttons, which is to filter data.
- The filter tabs need a title to help with usability
- The completion rate status bar could use the state colors defined in the conda-forge style guide (done=green, not solvable=orange, bot error=red). We still need to figure out colors for the 3 intermediate states, maybe some neutral colors would be best.

Here is an example of what this could look and feel like:

Migration details filtering


I will open a separate issue on this if necessary, as I think it shouldn't be a blocker the first version of this PR.

src/components/MigrationDetails/index.jsx
```diff
@@ -0,0 +1,254 @@
...
+      <Filters
+        counts={ORDERED.reduce((counts, [key]) =>
+          ({ ...counts, [key]: 0 || details[key]?.length }), {})}
+        filters={{ ...filters }}
+        onFilter={key => setState(prev => ({ ...prev, [key]: !prev[key] }))} />
+      {rows.length > 0 && <table>
+        <thead>
+          <tr>
+            <th style={{ width: 200 }}>Name</th>
+            <th style={{ width: 115 }}>Status</th>
```
jakirkham, line 205:
```suggestion
            <th style={{ width: 115 }}>PRs made</th>
```

jakirkham (commented):
Thanks for sharing today! 🙏

It's really nice to see how far this has come 👏

Thinking about the progress bar point earlier. Maybe we can just clarify the column header to make this a bit clearer? Proposed some wording below (though feel free to move if I got the wrong line and/or rephrase as needed)

beckermr (approved):
I'm more than happy to wait for other comments but for me lgtm!

jakirkham:
There are notes from today's meeting that we should follow up before finalizing

jaimergp (approved):
We are now in a mergeable state! Thanks for the hard work. I think the PR has addressed most comments now. There are still a few details we can polish in subsequent PRs (e.g. @GabrielaVives' suggestions), but this is good to go imo.

I'll merge next Monday if we don't get further comments :)

jaimergp:
Hm, in the meantime I noticed two little things in dark mode:

* The bar plot is almost invisible:

image

* The incidents table and sidebar borders are brighter than the other tables:

image

afshin:
> Hm, in the meantime I noticed two little things in dark mode:
> 
>     * The bar plot is almost invisible:
> 
> 
> image
> 
>     * The incidents table and sidebar borders are brighter than the other tables:
> 
> 
> image

These are both fixed, but I don't like the solution for the Chart colors. Because:

1. we use CSS variables for most colors, including the `ifm-color-primary` that is appropriate here
2. Chart.js can't accept CSS variables for bar constituent colors
3. `ifm-color-primar` can change at runtime if the user switches thems
4. There is no theme information passed into the `<StatusDashboar>`, so I need to use a `MutationObserver` on the `<html>` tag of the document and listen for changes to the `data-theme` attribute.

This works but seems inelegant to me. The better way would to have some way of being informed (via props or something) that the theme has changed. I am not convinced this is **The Right Way**™️  so if you have ideas, I would like to hear them.

beckermr:
@afshin i closed the degraded performance and major incident issues on the status report and now the new status page is crashing.

beckermr:
It says it cannot find the variable "issue"

afshin:
> It says it cannot find the variable "issue"

JavaScript strikes again. Should have picked Typescript 🤓

Thanks for letting me know!

beckermr:
Oh no! With no status issues open it displays an errant "0".

Screenshot 2024-03-09 at 12 29 00 PM

afshin:
I rushed this cause I was heading out the door. I’ll do a solid check of
incidents cases before Monday!

beckermr:
Have a good one!

jaimergp:
> I am not convinced this is **The Right Way**™️ so if you have ideas, I would like to hear them.

Maybe you can use https://docusaurus.io/docs/api/themes/configuration#use-color-mode?

jaimergp:
There are some conflicts because of #2093. I can help resolve them if you want, @afshin. Let me know and I'll push :)

SylvainCorlay:
> There are some conflicts because of https://github.com/conda-forge/conda-forge.github.io/pull/2093. I can help resolve them if you want, @afshin. Let me know and I'll push :)

That would be great!

jaimergp:
> > I am not convinced this is **The Right Way**™️ so if you have ideas, I would like to hear them.
> 
> Maybe you can use https://docusaurus.io/docs/api/themes/configuration#use-color-mode?

More dark mode visibility issues:

image

afshin:
> > I am not convinced this is **The Right Way**™️ so if you have ideas, I would like to hear them.
> 
> Maybe you can use https://docusaurus.io/docs/api/themes/configuration#use-color-mode?

I think I can use this, yes.  I'll give a try! Thanks @jaimergp!

And yes please, if merge conflicts arise that you know the right version of, please help! I'll make a comment in this thread when I have resolved the theme switching issue.

afshin:
The theme switching now uses the color modes. It still needs to  – unfortunately – compute the style value of the CSS color variables once at runtime (which is also why there are two new CSS variables that are *invariant* and define light and dark primary colors and are used to populate the respect dark/light `--ifm-color-primary` variables). This seems unavoidable. But the rest of it is cleaner without having to use a mutation observer for each change since I can rely on `useColorMode` for this, instead. The CSS has an explanatory note:

```css
:root {
  /*
    These two variables are defined in root and used at runtime for calculating
    the light mode and dark mode values of the primary color variable in order
    to populate `backgroundColor` prop `<UsageChart>`.
  */
  --ifm-color-primary-light-mode: #008478;
  --ifm-color-primary-dark-mode: #4db6ac;
  /* ... other stuff ... */
  --ifm-color-primary: var(--ifm-color-primary-light-mode);
  /* ... other stuff ... */
}

[data-theme="dark"] {
  --ifm-color-primary: var(--ifm-color-primary-dark-mode);
  /* ... other stuff ... */
}
```

---

Updating the colors of the details page, I would propose doing this as a follow-on.

jaimergp:
Thanks a bunch! Will merge by EOD.

SylvainCorlay:
:shipit:

jaimergp:
It's live! https://conda-forge.org/status

src/components/MigrationDetails/index.jsx
jakirkham: Making a similar change to the migration details page: https://github.com/conda-forge/conda-forge.github.io/pull/2113

jakirkham:
Looks like we lost the migration table's "Done" column (please see screenshot below). It is already in the details page

Adding it back in PR: https://github.com/conda-forge/conda-forge.github.io/pull/2114

Please let us know if anything else is needed

Screenshot 2024-03-11 at 2 09 20 PM

h-vetinari:
How long does it take for new migrations to get added to the status page, resp. must happen for it to be picked up? I cannot see a migration for libsentencepiece, 24h after merging https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5644...

jakirkham:
More likely this is a bot bug than a webpage bug

The webpage is [just reading]( https://github.com/conda-forge/conda-forge.github.io/blob/f14e650e93a449eb9fe6f01fba2a7e1e810dc777/src/constants.js#L88-L89 ) the [bot's data]( https://github.com/regro/cf-graph-countyfair/blob/master/status/regular_status.json )

The migration referenced doesn't show up there

Would suggest [raising a new bot issue]( https://github.com/regro/cf-scripts/issues/new )

PR #1950, opened 2023-04-24
update infrastructure.rst/compilers 

h-vetinari:
Reopen #1946 with more time for discussion, to ensure we don't accidentally change policy.

I made some small edits of oversights; diff compared to #1946 is below the fold:

<details>

```diff
diff --git a/src/maintainer/infrastructure.rst b/src/maintainer/infrastructure.rst
index 6e8d5bdc..c3f5bc35 100644
--- a/src/maintainer/infrastructure.rst
+++ b/src/maintainer/infrastructure.rst
@@ -318,21 +318,26 @@ roll out in a compatible manner to the ecosystem as feedstocks get rerendered.
 For such ABI-compatible upgrades, similar but looser principles apply:

 * The pins are similarly in the global pinning, see :ref:`globally_pinned_packages`.
-* We provide no support of any kind in terms of the long-term availability of a given compiler generation.
+* We provide no support of any kind in terms of the long-term availability of a given compiler version.
 * We generally provide notice in the form of an announcement when a compiler is going to be upgraded.
-* In general, our compilers on Linux and OSX are using very recent compilers, whereas
-  on windows, we generally use the last supported VS version.
+* Without promising any timelines, our compilers on Linux and OSX are normally
+  very recent; on Windows, we generally use the last supported VS version.

 Despite the lack of explicit support, we try to keep the compilers in their various versions
 working also outside of conda-forge, and even provide an easy way to install them
 (through the `compilers <https://github.com/conda-forge/compilers-feedstock>`_ feedstock).

-In more detail, our default compiler stack is made up very differently on each platform.
 More specifically, each compiler uses an _activation_ package that makes the difference
 between it being merely present in a build environment, and it being used by default.
 These will be installed when using ``{{ compiler('xyz') }}`` in ``meta.yaml``, where
 ``'xyz'`` is one of ``'c', 'cxx', 'fortran', 'cuda', 'rust', 'go-cgo', 'go-nocgo'``.

+Our default compiler stack is made up very differently on each platform; each platform
+has its own default compiler, with its own set of feedstocks that provide them. Due to
+historic reasons (the way compilers are integrated with their OS, and the amount of
+software written in them, etc.), the most impactful languages are C & C++ (though
+Fortran is considered part of the default, not least because GCC compiles all three).
+
 Linux (GCC):

 * [C, C++, Fortran] Activation: https://github.com/conda-forge/ctng-compiler-activation-feedstock/
```

</details>

OP from #1946 :
[...]

PR #2126, opened 2024-03-20
add announcement for upcoming stdlib migration

h-vetinari:
As discussed in https://github.com/conda-forge/conda-forge.github.io/issues/2102


h-vetinari:
pre-commit.ci autofix

news/2024-03-20-stdlib-migration.md
```diff
@@ -0,0 +1,47 @@
...
+
+After many discussions across conda-forge stakeholders, the solution we arrived
+at is the introduction of a new Jinja2-function `{{ stdlib("c") }}`, which
+reflects that a given recipe requires a C stdlib. Making this relationship
+explicit will make it easy to correctly reflect the requirement for newer
+stdlib versions per feedstock, as well as in our global pinning.
+
+Up until now, the stdlib was handled implicitly as part of the compiler stack.
+In order to allow this transition to happen, we need to introduce this function
+to essentially all compiled (i.e. not pure-Python) packages. This will be done
```
mfansler, line 22: Python shouldn't receive a special callout - this equally impacts all scripting languages with compiled library interfacing (e.g., Perl, R). I think the parenthetical could be dropped altogether - "all compiled packages" is sufficiently clear. Or perhaps "(i.e., non-noarch)".

Also maybe this should be "recipes" instead of "packages.

news/2024-03-20-stdlib-migration.md
```diff
@@ -0,0 +1,47 @@
...
+    - 2.17                       # [linux]
+  ```
+- if a feedstock sets `MACOSX_DEPLOYMENT_TARGET` in `conda_build_config.yaml`,
+  for example to 10.13 for `x86_64`, replace that section with the following
+  (note, this does _not_ apply to `MACOSX_SDK_VERSION`!):
+  ```
+  c_stdlib_version:              # [osx and x86_64]
+    - 10.13                      # [osx and x86_64]
+  ```
+  You should then also remove any line involving `__osx` from `meta.yaml`.
```
beckermr, line 45: Should we put this info in our knowledgebase entries on using compilers? It might already be there, IDK.

beckermr, line 45: >   You should then also remove any line involving `__osx` from `meta.yaml`.

Maybe it might be good to list an example of this? I don't know which lines you mean. Does this apply to say entries in run that have special osx run requirements?

jaimergp: Agreed that this would be a welcome addition to the knowledge base or compilers. We could link to the entry from the announcement so it doesn't get as long. Should help with discoverability!

h-vetinari: This was mainly trying to refer to (variations of) `- __osx >={{ MACOSX_DEPLOYMENT_TARGET }}  # [osx and x86_64]`

h-vetinari: > Should we put this info in our knowledgebase entries on using compilers?

Yeah, but that would need to build on https://github.com/conda-forge/conda-forge.github.io/pull/1950, which has been waiting for feedback/resolution. Perhaps I need to bring up this topic in the core call.

> We could link to the entry from the announcement so it doesn't get as long. Should help with discoverability!

Yeah, that sounds like a good idea.

beckermr: Got it. We should mention this explicitly.

beckermr: That PR doesn't have the new stdlib setup stuff in it for the knowledgebase that I can see.

h-vetinari: Yeah, that PR's about a year old. My point is that if we want to add stdlib handling to our compiler docs, I want to build on the existing clean-up rather than do two conflicting updates.

h-vetinari: > > We could link to the entry from the announcement so it doesn't get as long. Should help with discoverability!
> 
> Yeah, that sounds like a good idea.

I actually reconsidered this and don't necessarily agree anymore. The docs definitely need to be updated, but they should describe the new status quo, not a transition (or at least not for long, as it will become stale quickly).

The announcement (or tracking issue) seems like a reasonable place for me to describe the main mechanics of the transition, especially to the degree that it allows maintainers to do it themselves, without having to wait for the piggyback migrator.

jaimergp: My main concern was about the discoverability of that information, but the search features should reveal both, so no I have no strong feelings.

beckermr: We should clarify this point in the PR before it is merged.

h-vetinari: Definitely. Haven't had time to incorporate feedback here yesterday. I'm thinking to separate the announcement from the wider docs update though, which would also uncouple this from having to wait for #1950.

h-vetinari: So I only added a promise to update the docs for now, for two reasons:
* I want to see this working in the wild before we write docs
* I want #1950 to land first, but the announcement shouldn't have to wait for that.

Hope that's alright.

h-vetinari: Done now.

h-vetinari:
@conda-forge/core, any opposition to merging this as-is? I'll update the docs as soon as #1950 is settled.

beckermr (approved):

xhochy (approved):

h-vetinari:
Thanks for the reviews!

PR #2184, opened 2024-05-15
Update glossary

zklaus:
PR Checklist:

This PR is a minor update to the glossary, with style homogenization and a new entry for CDN.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -1,3 +1,7 @@
+---
+title: 'Glossary'
+---
+
 <a id="miscellaneous"></a>
 
 <a id="misc-glossary"></a>
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -1,79 +1,76 @@
----
-title: 'Glossary'
-sidebar_position: 27
----
-
-<a id="miscellaneous"></a>
-
-<a id="misc-glossary"></a>
-
-<a id="glossary"></a>
-
 # Glossary
 
-<a id="term-ABI"></a>
+List of frequently used terms and acronyms.
 
 ## ABI
 
-**A**pplication **B**inary **I**nterface. ABI is a document that comprehensively defines the binary system interface between applications and the operating system on which they run. [Learn More at Wikipedia](https://en.wikipedia.org/wiki/Application_binary_interface) or [pypackaging-native](https://pypackaging-native.github.io/background/binary_interface).
+Application Binary Interface. ABI is a document that comprehensively defines the binary system interface between applications and the operating system on which they run.
 
-<a id="term-CDT"></a>
+[Learn more](https://en.wikipedia.org/wiki/Application_binary_interface).
+
+## CDN
+
+Content Delivery Network. CDNs are geographically distributed networks of servers that mirror
+contents of a primary source. Having multiple servers offering the same content increases
+performance (reduced latency, higher download speeds) and availability.
+
+[Learn more](https://en.wikipedia.org/wiki/Content_delivery_network).
 
 ## CDT
 
-**C**ore **D**ependency **T**ree. Core Dependency Tree packages take care of the dependencies which are so close to the system that they are not packaged with conda-forge. A CDT package consists of repackaged CentOS binaries from the appropriate version, either 6 or 7 depending on user choice and platform. [Learn more](maintainer/knowledge_base.md#cdt-packages).
+Core Dependency Tree. Core Dependency Tree packages take care of the dependencies which are so close to the system that they are not packaged with conda-forge. A CDT package consists of repackaged CentOS binaries from the appropriate version, either 6 or 7 depending on user choice and platform.
+
+[Learn more](https://conda-forge.org/docs/maintainer/knowledge_base.html#cdt-packages).
 
 ## CLI
 
-**C**ommand **L**ine **I**nterface. A program in which commands are entered as text, one line at a time, for a computer to execute. This is done in the Miniforge Prompt in Windows, and in a terminal in macOS and Linux. `conda` is executed in a CLI. [Learn More](https://en.wikipedia.org/wiki/Command-line_interface).
-
-<a id="term-CI"></a>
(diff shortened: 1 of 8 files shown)


jaimergp:
The `<a id ...` anchors were there to re-export the old anchors from the old website, but we don't need to follow that format everywhere. Actually I lean towards moving away from the `#term-` prefix if possible (and only keep the old ones around for backwards compatibility with external sites linking to us, while using the new ones internally).

zklaus:
> The `<a id ...` anchors were there to re-export the old anchors from the old website, but we don't need to follow that format everywhere. Actually I lean towards moving away from the `#term-` prefix if possible (and only keep the old ones around for backwards compatibility with external sites linking to us, while using the new ones internally).

Makes sense to me. That means keeping the `<a ...` tags, but removing the `{#term-...}` tags and updating all links within the repo that refer to `glossary[.md]#term-` to remove the `term-` part, correct?

jaimergp:
Correct.

zklaus:
I think this is ready to go.

docs/glossary.md
```diff
@@ -1,8 +1,3 @@
----
-title: 'Glossary'
```
jaimergp, line 2: I think we need this line back (or push the `# Glossary` header to the absolute first line before the anchors). Otherwise the sidebar renders the title in lowercase.

docs/glossary.md
```diff
@@ -11,74 +6,116 @@ sidebar_position: 27
 
 # Glossary
 
+List of frequently used terms and acronyms.
+
 <a id="term-ABI"></a>
 
 ## ABI
 
-**A**pplication **B**inary **I**nterface. ABI is a document that comprehensively defines the binary system interface between applications and the operating system on which they run. [Learn More at Wikipedia](https://en.wikipedia.org/wiki/Application_binary_interface) or [pypackaging-native](https://pypackaging-native.github.io/background/binary_interface).
```
jaimergp, line 18: Why did you remove the `pypackaging` reference?

docs/glossary.md
```diff
@@ -11,74 +6,116 @@ sidebar_position: 27
...
+Content Delivery Network. CDNs are geographically distributed networks of servers that mirror the contents of a primary source. Having multiple servers offering the same content increases performance (reduced latency, higher download speeds) and availability.
+
+[Learn more](https://en.wikipedia.org/wiki/Content_delivery_network).
 
 <a id="term-CDT"></a>
 
 ## CDT
 
-**C**ore **D**ependency **T**ree. Core Dependency Tree packages take care of the dependencies which are so close to the system that they are not packaged with conda-forge. A CDT package consists of repackaged CentOS binaries from the appropriate version, either 6 or 7 depending on user choice and platform. [Learn more](maintainer/knowledge_base.md#cdt-packages).
+Core Dependency Tree. Core Dependency Tree packages take care of the dependencies that are so close to the system that they are not packaged with conda-forge. A CDT package consists of repackaged CentOS binaries from the appropriate version, either 6 or 7 depending on user choice and platform.
```
jaimergp, line 31: This part will get out of date soon. I think it's better to have this last sentence present an example.

> For example, in conda-forge, we have used repackaged CentOS 6 or 7 for some time.

docs/glossary.md
```diff
@@ -11,74 +6,116 @@ sidebar_position: 27
...
 
-<a id="term-CFEP"></a>
+[Learn More](https://en.wikipedia.org/wiki/Continuous_integration).
 
-## CFEP
+<a id="term-CLI"></a>
+
+## CLI
+
+Command Line Interface. A program in which commands are entered as text, one line at a time, for a computer to execute. This is done in the Miniforge Prompt in Windows, and in a terminal in macOS and Linux. `conda` is executed in a CLI.
```
jaimergp, line 55: I would rephrase the last two sentences to make sure people understand they are concrete examples.

docs/glossary.md
```diff
@@ -11,74 +6,116 @@ sidebar_position: 27
...
+[Learn More](https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/channels.html).
 
 <a id="term-Conda-package"></a>
 
 ## Conda package
 
-A conda package is a `.tar.bz2` or `.conda` archive that contains libraries, executable programs, data files and other components, as well as metadata under the `info/` directory. Its contents are unpacked in the installation prefix. [Learn More](<https://en.wikipedia.org/wiki/Conda_(package_manager)>).
+A conda package is a `.tar.bz2` or `.conda` archive that contains libraries, executable programs, data files and other components, as well as metadata under the `info/` directory. Its contents are unpacked in the installation prefix.
+
+[Learn More](<https://en.wikipedia.org/wiki/Conda_(package_manager)>).
```
jaimergp, line 73:
```suggestion
[Learn More](https://en.wikipedia.org/wiki/Conda_(package_manager)).
```

docs/glossary.md
zklaus: Snafu. It's back now.

docs/glossary.md
zklaus: Hm. To be honest, I am not thrilled about the entire explanation. What do we mean here with CLI? Imho, it can be either "the program that takes the command-line and launches the desired program", i.e. power shell, bash, etc.; or it can be the interface offered by programs such as `conda` including parameters etc.

When I say CLI, I usually mean the latter. The explanation here starts out with the former, but than kind of mangles the two; imho, the linked explanation on wikipedia refers to the second meaning.

zklaus: So I would most like to rewrite with the second meaning. Something along the lines of
```suggestion

## CLI

Command Line Interface. A textual interface of a program that allows the user to control the program with a single command line, which is entered in a command-line environment, such as the Miniforge Prompt on Windows, or a terminal on MacOS and Linux. An example for a program offering a CLI is `conda`.
```

zklaus:
*sigh* link checker failure due to ci network problems. @conda-forge-admin, please restart ci

jaimergp:
I don't think that works in this repo (but you can always close and reopen on your own). I issued a CI rerun, let's see.

jaimergp:
Thanks!

Issue #2132, opened 2024-03-26
Status page: query params instead of pretty URLs?


jaimergp:
@afshin was mentioning (please, extend this as necessary) that technically speaking, we shouldn't be using pretty URLs (`/status/migration/name`) for the migration details page, but instead query params (`/status/?migration=name`).

This would prevent the "404 flash" and would better comply with the "contract of page permanence".

afshin:
I think it would most likely look like `/status/migration?name={name}`, but yes, otherwise exactly as Jaime says!

jaimergp:
While we are at it, could we also add a `view={table,graph}` param?

afshin:
That is also pretty straightforward and sounds reasonable.

Issue #2137, opened 2024-03-26
[META] Status page improvements


jaimergp:
Just so we don't miss these pieces of feedback shared in #2090 and later:

- [x] https://github.com/conda-forge/conda-forge.github.io/issues/2135
- [x] https://github.com/conda-forge/conda-forge.github.io/issues/2132
- [x] [Contrast issues in migration details in dark mode](https://github.com/conda-forge/conda-forge.github.io/pull/2090#issuecomment-1988228279): https://github.com/conda-forge/conda-forge.github.io/pull/2239
- [x] Wheel-click or "Open link in new tab" action on any migration will take you back to the migrations overview, not the details: https://github.com/conda-forge/conda-forge.github.io/pull/2235
- [x] Incident titles do not render backticks: https://github.com/conda-forge/conda-forge.github.io/pull/2229
- [X] Migration details table loads up with all entries filtered out. Would be better to either show all by default, or at least print a message that says how many items are filtered out so users don't believe that there are no items: https://github.com/conda-forge/conda-forge.github.io/pull/2237
- [x] add [total children](https://github.com/conda-forge/conda-forge.github.io/issues/2137#issuecomment-2094059627) & sort feedstocks by it
- [x] [width](https://github.com/conda-forge/conda-forge.github.io/issues/2137#issuecomment-2106532161) of stacktrace foldout: https://github.com/conda-forge/conda-forge.github.io/pull/2230
- [x] display [paused](https://github.com/conda-forge/conda-forge.github.io/issues/2137#issuecomment-2113312542) migrators in status page
- [x] fix "not redirecting properly" when [using](https://github.com/conda-forge/conda-forge.github.io/issues/2137#issuecomment-2113319504) link to status page: https://github.com/conda-forge/conda-forge.github.io/pull/2235
- [x] version status errors need same formatting w/ details as migration ones: https://github.com/conda-forge/conda-forge.github.io/pull/2234
- [ ] revisit colour scheme for migrator stati (c.f. [here](https://github.com/conda-forge/conda-forge.github.io/issues/2137#issuecomment-2294793097))
- [ ] consider changing progress bar to "PRs merged", not "PRs opened" (c.f. [here](https://github.com/conda-forge/conda-forge.github.io/issues/2137#issuecomment-2303047226))

@GabrielaVives proposed these UX improvements in #2090:

[...]

PR #2235, opened 2024-07-19
Switch from dynamic URLs to query string parameters for migration details

afshin:
In order to prevent the 404 flash, we should stop using dynamically generated URLs for each migration details page and instead rely on a query string parameter. This PR fixes the 404 flash.

Fixes https://github.com/conda-forge/conda-forge.github.io/issues/2132

This issue is also mentioned in https://github.com/conda-forge/conda-forge.github.io/issues/2137 and fixes bullet points number 2 and number 4.

This PR (almost certainly) also fixes the issue [in this comment](https://github.com/conda-forge/conda-forge.github.io/issues/2137#issuecomment-2113319504).

**Screenshot:**

Screenshot 2024-07-19 at 14 36 26

Note the URL structure.

cc: @HaudinFlorence @jaimergp


jaimergp:
Awesome! While we are here, is it possible to add navigation params to:

- Table vs graph view
- Which categories are toggled on and off (Done, In PR, etc)

jaimergp:
> This issue is also mentioned in #2137 and fixes bullet points number 2 and number 4.
> 
> This PR (almost certainly) also fixes the issue [in this comment](https://github.com/conda-forge/conda-forge.github.io/issues/2137#issuecomment-2113319504).

Yea I can't reproduce those in this PR either.

afshin:
> Awesome! While we are here, is it possible to add navigation params to:
> 
>     * Table vs graph view
> 
>     * Which categories are toggled on and off (Done, In PR, etc)

Hm, should these be in the URL? I can see some value to it but I'm a little ambivalent.

I wouldn't combine it with this PR, we can just add these features to the backlog.

jaimergp:
> Hm, should these be in the URL? I can see some value to it but I'm a little ambivalent.

The issue I'm observing is that when I link to the migration page to new users, they are confused because initially there's no information displayed. They have to learn to click on the toggle buttons to display some stuff, and then they don't really know what statuses are representing. By having the categories in the URL I could link exactly to the part they need to pay attention to, instead of browsing / Ctrl+F'ing their package so much.

But yea, it can be done in a separate PR; only mentioned it in case it was low hanging fruit.

beckermr:
Why not just have everything toggled to on when the page loads? This seems more useful anyways.

jaimergp:
Another thing to consider, all the migration links that have been posted til now will start 404ing. Example https://deploy-preview-2235--conda-forge-previews.netlify.app/status/migration/aws_c_cal071. Is there a chance we can leave the navigation in place for a while so it redirects to the `?name=` equivalent?

jaimergp:
> Why not just have everything toggled to on when the page loads? This seems more useful anyways.

Yes, that's item 6 in https://github.com/conda-forge/conda-forge.github.io/issues/2137#issue-2209150215

afshin:
> Why not just have everything toggled to on when the page loads? This seems more useful anyways.

This is orthogonal to whether we put them in the URL, and I think you're correct. @HaudinFlorence is looking at the meta list and that's one of the issues she'll be handling.

afshin:
> Another thing to consider, all the migration links that have been posted til now will start 404ing. Example https://deploy-preview-2235--conda-forge-previews.netlify.app/status/migration/aws_c_cal071. Is there a chance we can leave the navigation in place for a while so it redirects to the `?name=` equivalent?

I am not 100% what the syntax is, but I believe Docusaurus supports URL rewriting for this purpose and that's a configuration rather than application code thing.

beckermr:
Right my point was that @jaimergp wants things in the URL so he can have them on when he sends links to confuse people less. If they are all on by default, then that would solve his problem partly and might aleviate the need for them in the links.

jaimergp:
I took a look at how we could reimplement the redirects but it doesn't work really well (double renders, more 404 flashing etc). Instead, I decided to catch the 404 page and print something more useful for the end user. They will have to click a link but at least it's not as confusing as "ha! it's gone!" message.

jaimergp:
Let's give this a try.

PR #2239, opened 2024-07-23
Change the colors of the filter buttons and change the migration details bar (colors and tooltips)

HaudinFlorence:
PR Checklist:

This PR aims at changing the colors of the filter buttons using `button--primary` and `button--secondary` classes as shown:
**LIGHT MODE**
before
Screenshot from 2024-07-23 16-22-50
now
Screenshot from 2024-07-23 16-23-33

**DARK MODE**
before
Screenshot from 2024-07-23 18-16-55

now
Screenshot from 2024-07-23 18-16-19

This PR aims at fixing task 3 and partially task 13 of issue https://github.com/conda-forge/conda-forge.github.io/issues/2137. It takes into account suggestions of this comment https://github.com/conda-forge/conda-forge.github.io/pull/2090#issuecomment-1979121823

cc @GabrielaVives @afshin


jaimergp:
Thanks! The buttons look perfect.

The "progress bar" still uses the same viridis colors as before, though, which has terrible visibility in dark mode:

image


I think Gabriela was suggesting to use tooltips to identify the different blocks of the progress bar in https://github.com/conda-forge/conda-forge.github.io/pull/2090#issuecomment-1979121823

beckermr:
The color coding is a usability plus in my book. People get a clear idea immediately and can guess at what the colors mean. I'd argue we should find better colors for dark mode as opposed to removing them.

jaimergp:
The concern raised by @GabrielaVives was:

> I have been through the multiple comments giving feedback about the filters and colors in the migration details section.
> There are several issues I would like to address:
> 
> - I think that the filtering buttons should have neutral colors rather than success or warning colors because their primary function is to categorize or sort data, not to convey a sense of achievement or caution. It will help promote clarity in this page where a lot of information is displayed, consistency, accessibility, and a focus on the primary function of the buttons, which is to filter data.
> - The filter tabs need a title to help with usability
> - The completion rate status bar could use the state colors defined in the conda-forge style guide (done=green, not solvable=orange, bot error=red). We still need to figure out colors for the 3 intermediate states, maybe some neutral colors would be best.

We need to be careful with color coding and contrast issues and it's usually tricky to find a palette that works well contrast wise _and_ doesn't run into color blindness issues. Hence why I assume Gabriela was suggesting different approaches that didn't rely exclusively on color coding.

The visibility in dark mode right now needs to be fixed one way or another. I don't feel strongly about one solution over the other as long as we keep accessibility concerns in mind. 

> @jaimergp I was thinking about implementing the changes on the bar and the tooltip in a different PR.

I think this time it'll be better to address both things in a single PR because we'll have to make a decision about how we approach the solution for both. Or at least, I'd like to see them together even if as a prototype, and then we can split in PRs. But the changes look reasonably small to tackle in a single PR.

beckermr:
Sounds good to me. I am not knowledgeable on these accessibility issues so I defer to you all on that.

HaudinFlorence:
@jaimergp Ok I will do everything in this same PR and will change the title of it.

jaimergp:
pre-commit.ci autofix

src/css/custom.css
```diff
@@ -39,6 +39,7 @@
   --ifm-color-info: #1976d2;
   --ifm-color-warning: #ef6c00;
   --ifm-color-danger: #d84315;
+
```
jaimergp, line 42:
```suggestion
```

Can we also remove the other formatting changes? I think we don't currently have an auto-formatter for the React bits so pre-commit is not catching it. We should add this in an another PR. Thanks!

src/pages/status/migration/styles.module.css
```diff
@@ -48,81 +47,29 @@
...
-  background-color: rgb(43, 93, 131);
-  border: 3px solid rgb(43, 93, 131, 0.4);
-  color: var(--ifm-color-secondary);
-}
-
-.migration_details_filter_in_pr_on {
-  background-color: rgb(43, 93, 131, 0.4);
-  border: 3px solid rgb(43, 93, 131);
-  color: var(--ifm-color-primary-dark);
+  background-color: var(--ifm-color-gray-500);
```
jaimergp, line 56:
```suggestion
  background-color: var(--ifm-color-emphasis-500);
```

This one is the same but reacts to dark mode to maintain contrast.

src/pages/status/migration/styles.module.css
```diff
@@ -48,81 +47,29 @@
...
-  background-color: rgb(47, 174, 110);
-  border: 3px solid rgb(47, 174, 110, 0.4);
-  color: var(--ifm-color-secondary);
-}
-
-.migration_details_filter_awaiting_pr_on {
-  background-color: rgb(47, 174, 110, 0.4);
-  border: 3px solid rgb(47, 174, 110);
-  color: var(--ifm-color-primary-dark);
+  background-color: var(--ifm-color-gray-600);
```
jaimergp, line 60:
```suggestion
  background-color: var(--ifm-color-emphasis-600);
```

src/pages/status/migration/styles.module.css
```diff
@@ -48,81 +47,29 @@
...
-  background-color: rgb(253, 227, 33);
-  border: 3px solid rgb(253, 227, 33, 0.9);
-  color: var(--ifm-color-primary-darker);
-}
-
-.migration_details_filter_awaiting_parents_on {
-  background-color: rgb(253, 227, 33, 0.4);
-  border: 3px solid rgb(253, 227, 33);
-  color: var(--ifm-color-primary-darker);
+  background-color: var(--ifm-color-gray-800);
```
jaimergp, line 64:
```suggestion
  background-color: var(--ifm-color-emphasis-800);
```

src/css/custom.css
HaudinFlorence: @jaimergp I will try to remove the formatting changes. I have tried to pay attention not use the keyboard shortcut to auto format the code but it is such a reflex.

HaudinFlorence:
@jaimergp If you add better suggestion for the tooltips text color and background color, please let me know. I thought that using the ones of the secondary button may be a solution, but I could not figure out how to get them from Infima.

HaudinFlorence:
pre-commit.ci autofix

src/pages/status/migration/index.jsx
```diff
@@ -111,13 +112,20 @@ function Bar({ details }) {
...
-              className={styles[`${prefix}${key.replace("-", "_")}`]}
-              style={{ flex: details[key].length }} key={index}></div>
-          ))}
+        {ORDERED.filter(([key]) => details[key]?.length).map(([key], index) => (
+          <>
+          <a
+            id={"migration-bar-element" + "-" + key}
+            className={styles[`${prefix}${key.replace("-", "_")}`]}
+            style={{ flex: details[key].length }}
+            key={index}
```
jaimergp, line 121:
```suggestion
            key={index}
            alt={key}
```

For screen readers, I think?

src/pages/status/migration/index.jsx
```diff
@@ -111,13 +112,20 @@ function Bar({ details }) {
...
+          ></a>
+         <Tooltip anchorSelect={"#migration-bar-element" + "-" + key}
+                  style={{color: "white", backgroundColor: "var(--ifm-color-gray-700)"}} place="top" className={styles.migration_details_bar_tooltip}>
```
jaimergp, line 124: Where's this class defined? I think we can put it in `styles.modules.css` with the `style` attributes you have added?

jaimergp:
Tooltips looking awesome! Just two comments.

src/pages/status/migration/index.jsx
HaudinFlorence: To be fair, I tried to use a class put in the `styles.modules.css` but it did not report the changes of colors so I decided to use the style attribute instead.

HaudinFlorence: I thought I removed the class attribute of the `Tooltip` component too but it is still there. If you know how to make it work with a class, I would be happy to know how to do it. Thanks

jaimergp: Let me give it a try!

jaimergp: Apparently we needed a more specific selector for this to work. See https://react-tooltip.com/docs/examples/styling#classes.

jaimergp (approved):

src/pages/status/migration/index.jsx
HaudinFlorence: > Apparently we needed a more specific selector for this to work. See https://react-tooltip.com/docs/examples/styling#classes.

I don't understand how it works with the `<style>` tag. I am having errors trying to translate the logics of the example in our specific case.

HaudinFlorence:
I am adding two gifs here to have an overview of the rendering in light and dark mode:
[Screencast from 24-07-2024 16:07:47.webm](https://github.com/user-attachments/assets/69f1d4d0-9882-48c1-b094-8f9e7cb56cd0)
[Screencast from 24-07-2024 16:10:52.webm](https://github.com/user-attachments/assets/7902c8aa-3194-4868-bf6f-a68570c846c4)

jaimergp:
Hi @HaudinFlorence, I presented your changes in the conda-forge core meeting today to see how folks felt about it. The reception was great; we just have two tiny comments:

- The difference between orange (warning) and red (error) in light mode is tricky to see for some folks / displays. I think we could try to increase the darkness in the red to make it more apparent. This can be done in a different PR.
- Would it be possible to have a little dot of color next to the filter button to quickly know which button refers to each progress bar slice? Something like this:

image

GabrielaVives:
> * Would it be possible to have a little dot of color next to the filter button to quickly know which button refers to each progress bar slice? Something like this:
> 
> image


After testing the filters, I believe that the filter icon on the left does not add value. The icon does not clearly indicate its purpose, and the hover or selection states of the buttons should suffice to convey their filtering actions, as is common in many applications and websites. Adding the icon suggests a lack of confidence in the UX of the filters and unnecessarily complicates the interface, distracting from the primary function of these buttons.

I recommend removing the filter icon and instead placing a color dot on the left side of each tab. This approach is more intuitive, as it maintains a consistent location for the color dot, regardless of the text length.

GabrielaVives:
> I am adding two gifs here to have an overview of the rendering in light and dark mode: [Screencast from 24-07-2024 16:07:47.webm](https://github.com/user-attachments/assets/69f1d4d0-9882-48c1-b094-8f9e7cb56cd0) [Screencast from 24-07-2024 16:10:52.webm](https://github.com/user-attachments/assets/7902c8aa-3194-4868-bf6f-a68570c846c4)

In my opinion, the behavior is good, but  the filters could be organised differently:

- Align all the filter buttons on the left
- Reduce the space between each button
- They are organised in the order of the progress bar above, but I don't think that brings any value because there is no visual correspondence. I would rather organise them based on their usage: for example, put the "done" button last.

beckermr:
right now I see this

Screenshot 2024-07-26 at 7 16 07 AM

I think we should remove the "filter symbol" and just have the dot centered vertically and left justified in the button. This is the same comment as @GabrielaVives made above

> I recommend removing the filter icon and instead placing a color dot on the left side of each tab. This approach is more intuitive, as it maintains a consistent location for the color dot, regardless of the text length.

HaudinFlorence:
> right now I see this
> 
> Screenshot 2024-07-26 at 7 16 07 AM
> I think we should remove the "filter symbol" and just have the dot centered vertically and left justified in the button. This is the same comment as @GabrielaVives made above
> 
> > I recommend removing the filter icon and instead placing a color dot on the left side of each tab. This approach is more intuitive, as it maintains a consistent location for the color dot, regardless of the text length.

I was planning to remove it, but was maybe expecting more feedback on this suggestion before doing it. So let's remove it.

beckermr:
Ah great! Let's go ahead indeed!

HaudinFlorence:
> > I am adding two gifs here to have an overview of the rendering in light and dark mode: [Screencast from 24-07-2024 16:07:47.webm](https://github.com/user-attachments/assets/69f1d4d0-9882-48c1-b094-8f9e7cb56cd0) [Screencast from 24-07-2024 16:10:52.webm](https://github.com/user-attachments/assets/7902c8aa-3194-4868-bf6f-a68570c846c4)
> 
> In my opinion, the behavior is good, but the filters could be organised differently:
> 
> * Align all the filter buttons on the left
> * Reduce the space between each button
> * They are organised in the order of the progress bar above, but I don't think that brings any value because there is no visual correspondence. I would rather organise them based on their usage: for example, put the "done" button last.

@GabrielaVives Thanks for this comment. I implemented the first 2 points. For the third one, I am not sure what is the best order for a reorganisation the button. Any detailed suggestion is welcome.

HaudinFlorence:
pre-commit.ci autofix

HaudinFlorence:
The last commits are attempts to improve the colors of the migrations bars and dots. I think if there is no further suggestions and objections to the changes performed that all the mentioned issues related to the filters buttons have been addressed and reported elsewhere in a new issue.

beckermr:
This looks great and I think we should merge. We can iterate more in future PRs. Thank you so much!

PR #2245, opened 2024-07-29
Fix some contrast issues in the filter buttons

jaimergp:
PR Checklist:

Follow up to #2239, to remove the red/orange border which didn't play well with some greens:

image

image

Now we use dark-mode aware blacks or whites, depending on the color at hand.

### DIFF
--- a/src/pages/status/migration/index.jsx
+++ b/src/pages/status/migration/index.jsx
@@ -159,7 +159,6 @@ function Breadcrumbs({ children }) {
 }
 
 function Filters({ counts, filters, onFilter }) {
-  const icon = styles.migration_details_filter_icon;
   return (
     <div className={styles.migration_details_filter}>
       {ORDERED.map(([key, title], index) => {
@@ -176,13 +175,16 @@ function Filters({ counts, filters, onFilter }) {
           onClick={() => onFilter(key)}>
           {filters[key] ?
             <span className={[
-              styles[`${base}_on`],
-              styles.migration_details_filter_dot_on
+              styles[`${base}_hidden`],
+              styles.migration_details_filter_dot,
+              styles[`${base}_dot`],
             ].join(" ")}></span>
             :
             <span className={[
               styles[base],
-              styles.migration_details_filter_dot].join(" ")}>
+              styles.migration_details_filter_dot,
+              styles[`${base}_dot`],
+            ].join(" ")}>
             </span>
           }
           <div className={styles.migration_details_filter_title_container}>
--- a/src/pages/status/migration/styles.module.css
+++ b/src/pages/status/migration/styles.module.css
@@ -52,7 +52,7 @@
   background-color: var(--ifm-color-success);
 }
 
-.migration_details_filter_done_on {
+.migration_details_filter_done_hidden {
   background-color: var(--ifm-color-success);
   opacity: 0.6;
 }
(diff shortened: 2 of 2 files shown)


beckermr:
Could we remove the boarder all together or does that create other issues?

HaudinFlorence:
> Could we remove the boarder all together or does that create other issues?

To give an idea, without any border for the dost, it would look like this:
Screenshot from 2024-07-29 15-07-23
Screenshot from 2024-07-29 15-07-32

HaudinFlorence:
Another alternative of what is proposed in the current PR and what was proposed in https://github.com/conda-forge/conda-forge.github.io/pull/2239, is to remove the red border only. It would look like this:
Screenshot from 2024-07-29 15-16-24

HaudinFlorence:
Another alternative of what is proposed in the current PR and what was proposed in https://github.com/conda-forge/conda-forge.github.io/pull/2239, is to remove the red border only. It would look like this:
Screenshot from 2024-07-29 15-16-24
Screenshot from 2024-07-29 15-16-30

jaimergp:
The "In PR" color on the light background is almost invisible, so that's why it's special cased in the CSS. This is the best compromise of accessibility / simplicity I could find.

HaudinFlorence:
> The "In PR" color on the light background is almost invisible, so that's why it's special cased in the CSS. This is the best compromise of accessibility / simplicity I could find.

Ok. That's right.

beckermr:
I like it without borders if that is easier.

HaudinFlorence:
Working without border would probably require to change the gray colors of the `in PR` , `awaiting PR` and `awaiting parents` categories, because in the present situation one of the gray is too close to the background.

HaudinFlorence:
Applying the same border color chosen to fix the light grey `in PR` but for all dots s also a possibility. My concern is that using different colors for borders gives a kind of the optical illusion that the dots have different sizes. I also personnally prefer to have something some homogeneous.

HaudinFlorence:
> Applying the same border color chosen to fix the light grey `in PR` but for all dots s also a possibility. My concern is that using different colors for borders gives a kind of the optical illusion that the dots have different sizes. I also personally prefer to have something some homogeneous.

After testing this solution, it is not satisfactory in dark mode as shown hereunder
Screenshot from 2024-07-29 21-57-47
Screenshot from 2024-07-29 21-58-11

src/pages/status/migration/index.jsx
```diff
@@ -176,13 +175,16 @@ function Filters({ counts, filters, onFilter }) {
...
+              styles[`${base}_hidden`],
+              styles.migration_details_filter_dot,
+              styles[`${base}_dot`],
             ].join(" ")}></span>
             :
             <span className={[
               styles[base],
-              styles.migration_details_filter_dot].join(" ")}>
+              styles.migration_details_filter_dot,
+              styles[`${base}_dot`],
```
HaudinFlorence, line 186: @jaimergp I may be wrong but it seems that line 185 and 186 are refering to the same class.

jaimergp:
I'd stick with the proposed solution if we cannot improve it for now. We can always refine later in further PRs.

HaudinFlorence:
> I'd stick with the proposed solution if we cannot improve it for now. We can always refine later in further PRs.

@jaimergp That's ok for me as it is in this PR.

Issue #2255, opened 2024-08-04
RFC: Drop PyPy


h-vetinari:
It gives me no pleasure to write this, but the signs have been mounting for some time.

Since the [introduction](https://conda-forge.org/blog/2020/03/10/pypy/) of PyPy in conda-forge ~4.5 years ago, we've migrated for 3.6, 3.7, 3.8 & 3.9, introduced windows support halfway through, and made hundreds of packages available pre-compiled for PyPy, or even PyPy-compatible upstream in the first place.

The lion's share of the work there was done by @mattip, one of the PyPy devs, who was truly tireless in getting this effort off the ground. Huge kudos for this work! But unsurprisingly, no single person can keep this up forever, and since the rest of cf/core does not have a lot of PyPy expertise (much less spare time), it all came down predominantly to Matti. When signs of fatigue showed up, I proposed to drop the least common architectures for PyPy
https://github.com/conda-forge/conda-forge.github.io/blob/14deb66144d112a52d38b9ff62f6c6673a105fb7/community/minutes/2023-10-04.md?plain=1#L55-L58

This was rejected at the time, but the situation hasn't changed since then. The linked [issue](https://github.com/conda-forge/pypy3.6-feedstock/issues/104) is a year old, but not much has changed since then. In particular, we have never migrated for PyPy 3.10, which is becoming an issue because many NEP-29 packages are starting to drop Python 3.9 support, and more importantly, we de-scoped PyPy from the numpy 2.0 migration [because](https://github.com/conda-forge/scipy-feedstock/pull/274#issuecomment-2092041382) we concluded
> @mattip: Maybe it is time to drop PyPy from future migrations. Packages that have been built should continue to work, so current users would not be suddenly cut off, and maintainers would not have to deal with the extra headache. I don't see a PyPy3.10 migration happening, and 3.9 is being dropped in the scientific python ecosystem. 

When this briefly came up in a core call a few months ago (not minuted AFAIR), dropping PyPy was already considered an option. I think it's time to make this official. If desired, we could draw this out until we close the numpy 2.0 migration, but in the meantime I think we should announce that we're sunsetting support and stop the PyPy migrator.

No matter the outcome, again a huge thanks to @conda-forge/pypy3-6 @conda-forge/pypy-meta, and @mattip & @isuruf in particular! 👏 🫡

hmaarrfk:
[...]

PR #2259, opened 2024-08-08
Add announcement for sunsetting PyPy

h-vetinari:
Draft for #2255, based on (my understanding of) the core call of today.

CC @conda-forge/core

### DIFF
--- a/news/2024-08-10-sunsetting-pypy.md
+++ b/news/2024-08-10-sunsetting-pypy.md
@@ -1,0 +1,37 @@
+# Sunsetting PyPy support
+
+Conda-forge introduced [support](https://conda-forge.org/blog/2020/03/10/pypy/)
+for [PyPy](https://pypy.org/index.html) as an alternative implementation of
+Python about 4.5 years ago.
+
+In that time we have worked hard together with developers from PyPy to provide
+easily installable pre-compiled builds of the most common libraries also for PyPy.
+
+As a very positive side-effect, the infrastructure of conda-forge is now fully
+equipped to deal with alternative implementations of the Python interpreter,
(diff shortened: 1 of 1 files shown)


jaimergp:
The branch name 😂  The content has the information we need with a very gentle tone that I'm sure folks will appreciate. Thanks for that.

For long-ish announcements I recommend adding a first paragraph with the TLDR, then adding a `<!-- truncate -->` comment in its own line, and then we can write as much as we want. This way the `news/` page is quicker to look, and details can be obtained in the full, single-page view of the announcement.

jaimergp:
LGTM. Tagging @mattip for visibility.

news/2024-08-10-sunsetting-pypy.md
```diff
@@ -0,0 +1,44 @@
+# Sunsetting PyPy support
+
+TL;DR: We are planning to remove all PyPy builds in conda-forge in a few weeks,
```
mattip, line 3: The original makes it sound like packages will be removed, I think the intention is to stop building new PyPy packages, correct?
```suggestion
TL;DR: We are planning to stop building new releases of packages for PyPy and
remove PyPy from conda-forge feedstock recipes in a few weeks,
```

mattip:
Thanks also to the conda-forge team who put a lot of effort into making this work for as long as it did!

news/2024-08-10-sunsetting-pypy.md
h-vetinari: This is a good clarification, thanks. I'm just too packaging-brained, where "builds" are obviously new (because we don't ever remove artefacts).

jaimergp (approved):

PR #2262, opened 2024-08-12
Add a table for the paused migrators category

HaudinFlorence:
PR Checklist:

This PR aims at fixing task 9 of meta issue https://github.com/conda-forge/conda-forge.github.io/issues/2137. It adds a new category for paused migrations and add a new corresponding table.
It uses recent PR https://github.com/regro/cf-scripts/pull/2886

Screenshot from 2024-08-12 16-52-30


src/components/StatusDashboard/patch.diff
jaimergp, line 1: What is this file for?

src/components/StatusDashboard/current_migrations.jsx
```diff
@@ -100,6 +102,16 @@ export default function CurrentMigrations({ onLoad }) {
             sort={state.sort.closed}
           />
         </table>
+        <table className={styles.migrations_table}>
+          <TableContent
+            collapsed={state.collapsed.paused}
+            name="Paused migrations"
+            resort={resort("paused")}
+            rows={paused}
```
jaimergp, line 110:
```suggestion
            rows={paused || []}
```

To avoid the `...` in the count, when it should be zero, I think?

src/components/StatusDashboard/patch.diff
HaudinFlorence: Sorry. That's a left over. I will remove it.

src/components/StatusDashboard/current_migrations.jsx
jaimergp: Hm, that didn't work. It might be somewhere else in the code then. My concern here is that we display `...` even when the (empty) file loaded. And we should distinguish from "waiting for file" (signified by `...`) and "file was received but is empty" (this should be `0`).

HaudinFlorence: Ok I will try to fix that.

src/components/StatusDashboard/current_migrations.jsx
```diff
@@ -100,21 +102,32 @@ export default function CurrentMigrations({ onLoad }) {
...
+  console.log('name:', name);
   if (redirect) return (<Redirect to={redirect} replace={false} push={true} />);
   return (
     <>
       <thead>
         <tr onClick={select}>
           <th colSpan={8} className={collapsed ? styles.collapsed : undefined}>
             {name}{" "}
-            <span className="badge badge--secondary">{rows.length || "…"}</span>
+            <span className="badge badge--secondary">{name!=="Paused migrations" ? rows.length || "..." : rows.length || "0"}</span>
```
jaimergp, line 130: Are we special casing "Paused migrations" here? I would expect the code to be something like:

- While fetching the remote JSON, the badge says `…` (note this is the three period symbol, not three periods one after another like `...`; this should be reverted).
- Once fetched, we calculate the `rows.length` and then report the number.

Is this too late in the code to check for rows being fetched? 🤔

src/components/StatusDashboard/current_migrations.jsx
```diff
@@ -100,21 +102,32 @@ export default function CurrentMigrations({ onLoad }) {
...
+          />
+        </table>
       </div>
     </div>
   );
 }
 
 function TableContent({ collapsed, name, resort, rows, select, sort }) {
   const [redirect, setState] = useState('');
+  console.log('name:', name);
```
jaimergp, line 122:
```suggestion
```

src/components/StatusDashboard/current_migrations.jsx
HaudinFlorence: > Is this too late in the code to check for rows being fetched? 🤔

I don't know

HaudinFlorence: Some logics was added to address the mentioned point.

src/components/StatusDashboard/current_migrations.jsx
```diff
@@ -59,9 +61,10 @@ export default function CurrentMigrations({ onLoad }) {
       return { ...prev, collapsed: updated };
     });
   useEffect(fetchContent(onLoad, setState), []);
-  const { closed, longterm, regular } = state;
-  const total = closed.length + longterm.length + regular.length;
-  return (
+  const { closed, longterm, paused, regular } = state;
+  const total = closed.length + longterm.length + paused.length + regular.length;
+  const fetched = total > 0;
```
jaimergp, line 66: Cool trick! I had assumed that the four tables are fetched in parallel but looks like the way `fetchContent` is written, it's blocking til all four JSONs have been obtained? Not a problem for this PR, though.

jaimergp (approved):

h-vetinari:
Great to see this happening! FYI, as of https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6292, you'll have a paused migrator to play with. ;-)

src/components/StatusDashboard/current_migrations.jsx
afshin: Because those requests are to static assets and should come back quickly, the fetching is hidden behind one single promise. They are *fetched* in parallel but rendering awaits all of the promises resolving before it proceeds.

jaimergp (changes requested):
Oops, [this logic](https://github.com/regro/cf-scripts/blob/e0ab929aa458bc8378c7e08ab484abe9ca0a74e9/conda_forge_tick/status_report.py#L432-L441) I added is not correct because long-running ones won't ever make it to the "is it paused" check. I need to adjust a couple things in the bot. I'll do it after lunch :D

jaimergp:
I submitted a couple fixes to the backend (https://github.com/regro/cf-scripts/pull/2932, https://github.com/regro/cf-graph-countyfair/pull/33) but the paused pypy migrator still doesn't show up as paused... I'll need to look into it.

jaimergp:
Yay, thanks to @beckermr's https://github.com/regro/cf-scripts/pull/2940, the [paused migrators file](https://github.com/regro/cf-graph-countyfair/blob/master/status/paused_status.json) is now populated, which means that the table works:

image

The only thing I'm missing is some sort of indicator in the [details page](https://deploy-preview-2262--conda-forge-previews.netlify.app/status/migration/?name=pypy38), like an admonition that says the migrator is paused or closed:

> [!NOTE]
> This migration was recently closed.

Or

> [!NOTE]
> This migration is currently paused.

The migration file does not include the information needed, but we can check whether the migrator name is in one of the already fetched closed/paused lists.

Issue #2272, opened 2024-08-26
Infrastructure update for Channel Cloning/CDN sync


morremeyer:
### Your question:

Hey everyone, 

we (Anaconda) have been working on upgrades for the channel cloning/CDN sync infrastructure of anaconda.org, which is is used for conda-forge.

## Time planning

We are planning to switch to the new infrastructure tomorrow, 2024-08-27 at 12:30 UTC.
**No downtime is expected for users**.

Due to the amount of changes on the conda-forge channel, the initial sync after the switch to the new infrastructure will be a bit slower than the usual syncs - it is currently running in a "dry run" mode on the new infrastructure already to keep its caches up to date, and I expect the initial sync on the new infrastructure to take ~30 minutes. 

We have already implemented this change for the r and anaconda channels to ensure that our maintenance process and the new infrastructure are working as desired.

## What's changing?

* The codebase has been updated to make the cloning more fault-tolerant and reliable
* The new infrastructure is more tightly integrated into our observability stack, enabling us to detect problems earlier and more granularly (e.g. a specific package that fails to clone instead of just a general cloning error)ö
* We're adding a `last-updated.json` file at the channel root, which contains the last update time both as UNIX and ISO8601 timestamps. We are using this directly to monitor the updates end-to-end. Check it out for the [anaconda](https://conda.anaconda.org/anaconda/last-updated.json) channel if you're interested!

## Other notes

This topic was discussed in the conda-forge meeting on 2024-08-21 and it was decided to bring it up here to raise awareness.

If there's any questions, please ask!

h-vetinari:
Thank you for all your work on this! 🙏

morremeyer:
The initial clone on the new infrastructure started on 12:34 UTC and is currently ongoing. 

I'll update here again when it is done.

morremeyer:
The initial clone finished uploading and purging the Cloudflare cache at 13:18, we are now enabling the scheduled runs.

morremeyer:
We have improved the update frequency to be under 20 minutes now.

We will continue to look into improvements this week, as we already have ideas for that.
For now, we will let the process run as-is to get more insights into where the performance can be further improved.

jezdez:
Neat to be able to check https://conda.anaconda.org/conda-forge/last-updated.json now :)

jaimergp:
Cool, I'll update the status page!

jakirkham:
Am happy this went so smoothly! 😄

[...]

PR #2278, opened 2024-08-27
Update status badges for Anaconda.org

jaimergp:
PR Checklist:

Comes from https://github.com/conda-forge/conda-forge.github.io/issues/2272#issuecomment-2312999325

### DIFF
--- a/src/components/StatusDashboard/repos_and_bots.jsx
+++ b/src/components/StatusDashboard/repos_and_bots.jsx
@@ -57,12 +57,12 @@ function Image({ alt, link, children }) {
 }
 
 function CDNStatus() {
-  const [{ minutes, status }, setState] = useState({ minutes: 0, status: "…" });
+  const [{ minutes, status }, setState] = useState({ minutes: "…", status: "…" });
   useEffect(() => {
     void (async () => {
       try {
-        const response = await (await fetch(urls.repos.cdn.api)).text();
-        const updated = new Date(response.trim()).getTime();
+        const response = await (await fetch(`${urls.repos.cdn.api}?bustcache=${Date.now()}`)).json();
+        const updated = new Date(response.timestamp.iso8601).getTime();
         const delta = (new Date()).getTime() - updated;
         const status = delta < OPERATIONAL_WINDOW ? "operational" :
           delta < DEGRADED_WINDOW ? "degraded" : "major outage";
--- a/src/constants.js
+++ b/src/constants.js
@@ -126,8 +126,8 @@ export const urls = {
       },
     ],
     cdn: {
-      api: "https://s3.amazonaws.com/conda-static.anaconda.org/conda-forge/last-updated",
-      link: "https://conda-static.anaconda.org/conda-forge/rss.xml",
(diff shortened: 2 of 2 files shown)


src/components/StatusDashboard/repos_and_bots.jsx
```diff
@@ -57,12 +57,12 @@ function Image({ alt, link, children }) {
...
 
 function CDNStatus() {
-  const [{ minutes, status }, setState] = useState({ minutes: 0, status: "…" });
+  const [{ minutes, status }, setState] = useState({ minutes: "…", status: "…" });
   useEffect(() => {
     void (async () => {
       try {
-        const response = await (await fetch(urls.repos.cdn.api)).text();
-        const updated = new Date(response.trim()).getTime();
+        const response = await (await fetch(`${urls.repos.cdn.api}?bustcache=${Date.now()}`)).json();
```
jaimergp, line 64: Weird that we need this `bustcache` trick again (also needed in the [packages page for the RSS stuff](https://github.com/conda-forge/conda-forge.github.io/blob/8399abe74226fc690d9b8e37e4083c9039452158/src/components/Packages/index.jsx#L82-L87)). Is this a `conda.anaconda.org` misconfiguration? We don't need it for e.g. `raw.githubusercontent.com`.

cc @jezdez @morremeyer

beckermr:
the preview says the CDN is down. is that true?

morremeyer:
@beckermr This is how it looks for me right now: 
image

src/components/StatusDashboard/repos_and_bots.jsx
morremeyer: @jaimergp I'll look into how we can improve that this week. We explicitly purge all updated files from the Cloudflare cache at the end of each cache run.

However, we do also set a `Cache-Control: public, max-age=1200` header in responses so that clients don't always download the files anew. 

I expect that this leads to the effect that the status does not update: If the browser requests the file when the last update was 19 minutes ago, it will serve it from cache for the next 20 minutes, so it will show the last update being 39 minutes ago even though there should have been another update in the meantime.

morremeyer:
@beckermr I think you most likely ran into the issue I just described in https://github.com/conda-forge/conda-forge.github.io/pull/2278#discussion_r1733318315.

If in doubt (until we have improved that), hard refresh the page.

beckermr (approved):

src/components/StatusDashboard/repos_and_bots.jsx
jaimergp: Maybe the `max-age` for these status files (`last_updated.json`, `rss.xml`, etc) should be smaller, like a minute or something? I'm busting the browser cache with these `?bustcache=<timestamp>` hacks but ideally we wouldn't need them.

morremeyer: Yup, my thoughts exactly.

jezdez: I'd stick with the work-around for now.

jezdez (approved):

Issue #629, opened 2018-08-17
abort builds with [skip ci]/etc


dougalsutherland:
Even though CircleCI supposedly supports `[skip ci]`, I still usually see builds go through when I add that; not sure what I'm doing wrong, if they dropped it, or if something is misconfigured on our end. We could have the build scripts, maybe `ff_ci_pr_build.py`, also check the commit message and stop if it's in there.

Also, since CircleCI doesn't support `[skip circle]` and don't seem to want to (https://github.com/conda-forge/conda-forge.github.io/issues/45#issuecomment-198060065), we could more-or-less implement that here. Same for `[skip travis]`. Or we could implement `[skip mac]` or something to still work when we use CircleCI for Mac builds also, or similar.

Drawback: it would prevent you from manually running builds of commits that shouldn't actually have been skipped, unless we add something to do that.

ocefpaf:
> Drawback: it would prevent you from manually running builds of commits that shouldn't actually have been skipped, unless we add something to do that.

Not a big deal IMO, the advantages do outweigh this disadvantage.

I'm not sure how hard would be to implement that but I like the idea of a `[skip mac]` and `[skip linux]`.

dougalsutherland:
I think all we need to do is have the build script run a `git` command to check the current commit. Shouldn't be too difficult; I might give it a shot this weekend.

kynan:
Any update on this? Skipping CI for Azure [does not seem to work](https://github.com/conda-forge/staged-recipes/issues/1148#issuecomment-817121438).

jaimergp:
I added:

- https://github.com/conda-forge/staged-recipes/pull/27765
- https://github.com/conda-forge/conda-smithy/pull/2077

With those merged and released we can close this.

Issue #498, opened 2017-12-07
Docs for [ci skip] et al


CJ-Wright:
It might be nice to have docs on the commit message based CI skipping (and maybe a discussion about it's being temperamental?

jakirkham:
Also issue ( https://github.com/conda-forge/staged-recipes/issues/1148 ) is relevant. Note that it is a non-actionable issue just tracking this info. Would be happy to close it and have that content absorbed into a doc section somewhere instead.

jaimergp:
This can be closed. See https://github.com/conda-forge/conda-forge.github.io/pull/2297.

PR #2297, opened 2024-09-14
Update ci skip documentation in infrastructure.md

wshanks:
PR Checklist:

I noticed that when you use the bot to do non-packaging changes like update maintainer it only uses `[ci skip]`. So is it safe for the documentation to suggest only `[ci skip]` now?

Also, are the linked issues still relevant? https://github.com/conda-forge/conda-forge.github.io/issues/629 does not seem useful to me. It talks about some nuances with Travis, Circle, and Azure, but I am not sure what to take away as actionable. All my feedstocks only use Azure now. I think https://github.com/conda-forge/staged-recipes/issues/1148 is similarly linked for nuances between CI systems.

Also, I think #498 can be closed since skipping CI is already in the docs.

### DIFF
--- a/docs/maintainer/infrastructure.md
+++ b/docs/maintainer/infrastructure.md
@@ -445,7 +445,7 @@ repo [README.md](https://github.com/conda-forge/webservices-dispatch-action) for
 
 ### Skipping CI builds
 
-To skip a CI build for a given commit, put `[ci skip] ***NO_CI***` in the commit message.
+To skip a CI build for a given commit, put `[ci skip]` in the commit message.
 
 :::note[Related links]
 


jaimergp:
There have been several iterations in Azure (see https://github.com/Microsoft/azure-pipelines-agent/issues/1270), so we had to combine different "skipper" formulas to make sure we were not triggering CI services in large migrations (e.g. [admin-migrations](https://github.com/conda-forge/admin-migrations/blob/e5711acbfbe64b797ca790b62d5610ad3a4a8466/README.md?plain=1#L19)).

So it's one of those things that might not hurt to keep around "just in case".

wshanks:
Hmm, https://github.com/microsoft/azure-pipelines-agent/issues/1270 is interesting. It highlights that the skipper phrases only work with Azure Pipelines for merge commits, not for PRs. For PRs, the CI is still triggered. Here is a recent PR from the bot that uses `[skip ci]`: https://github.com/conda-forge/adversarial-robustness-toolbox-feedstock/pull/50. We can see that Azure Pipelines still runs on https://github.com/conda-forge/adversarial-robustness-toolbox-feedstock/pull/50/commits/3a19536add1e5dda1fa27d18fc3c1af8efc0141c while the merge commit https://github.com/conda-forge/adversarial-robustness-toolbox-feedstock/commit/0be9377b31d76a5720917432673252f3ef125611 does not. The GitHub Actions cf linter does not run on the PR though. As a test, I opened a PR using `***NO_CI***` in the title and commit message in https://github.com/conda-forge/ibm-platform-services-feedstock/pull/63. In this case, both the Azure Pipelines and GitHub Actions jobs ran. This agrees with https://github.com/conda-forge/staged-recipes/issues/1148#issuecomment-817125955 that you linked to.

So I don't see a case where `***NO_CI***` does something that `[skip ci]` does not do. I will leave it though if you think that is best. If we don't remove the `***NO_CI***` suggestion, we can change this PR to mention that the phrase only works on merge commits and not PRs and remove the old links. Is that what you would prefer?

I am not sure if CI running on PRs is what conda-forge wants or just a consequence of Azure Pipelines' design decision. conda-forge could add its own rule into its Azure Pipelines templates to skip PRs to save some CI time like numpy did (https://github.com/numpy/numpy/pull/21879#issuecomment-1217908810).

jaimergp:
> The GitHub Actions cf linter does not run on the PR though. 

Uhh yea, that's not great. @beckermr, what do you think about this? Does it interact with the "mergeability" checks?

> I am not sure if CI running on PRs is what conda-forge wants or just a consequence of Azure Pipelines' design decision. conda-forge could add its own rule into its Azure Pipelines templates to skip PRs to save some CI time like numpy did ([numpy/numpy#21879 (comment)](https://github.com/numpy/numpy/pull/21879#issuecomment-1217908810)).

I think it's mostly an AZP limitation. [This logic you linked](https://github.com/numpy/numpy/pull/21879/files#diff-7915b9b726a397ae7ba6af7b9703633d21c031ebf21682f3ee7e6a4ec52837a5) to could be added to the conda-smithy templates to mitigate this surprising behaviour, but for sure we should document this clearly now so it's not buried in an issue somewhere.

beckermr:
I believe this is the intended behavior for the linter. If it sees `[ci skip]` it skips ci.

jaimergp:
Thanks @wshanks, ended up adding https://github.com/conda-forge/conda-smithy/pull/2077 and https://github.com/conda-forge/staged-recipes/pull/27765 for Azure via your suggestion.

docs/maintainer/infrastructure.md
```diff
@@ -445,7 +445,7 @@ repo [README.md](https://github.com/conda-forge/webservices-dispatch-action) for
 
 ### Skipping CI builds
 
-To skip a CI build for a given commit, put `[ci skip] ***NO_CI***` in the commit message.
+To skip a CI build for a given commit, put `[ci skip]` in the commit message.
 
 :::note[Related links]
```
jaimergp, line 450: We can remove this admonition now, I think.

jakirkham:
Originally `[ci skip]` and `[skip ci]` worked for all CI providers (though I think Travis only liked one of those and forget what it was) and we added it to the linter

As we have moved from the original setup of Travis, CircleCI and AppVeyor through various additions and removals we have reached the current state of Azure, Travis and GitHub Actions (leveraged directly or using some other self-hosted runner solution)

Have forgotten some of the nuances of the different systems in between and how they handled skip notes. That said, `[ci skip] and `[skip ci]` are pretty standard on most CI providers. So we have tried to make sure that works somehow. Thought Azure did some work on supporting this somewhere, but it has been a while since I looked and there are probably caveats

Since sometimes CI providers have not consistently done what we have wanted. For a time we added logic to fast cancel jobs in certain conditions (not the latest build for a PR or commit, recipe has a `skip: true` condition, PR doesn't merge cleanly into base branch, etc.) as well as test on merge commits consistently. Think as systems changed we dropped that logic. We could bring something like that back if it would improve user experience

beckermr:
Thanks @jakirkham. We just merged a pr to fast finish prs with skip messages for azure in smithy. To be honest it seems like we should just keep the logic around no matter what given the turnover over the years.

PR #2385, opened 2024-11-24
lower_case_true

raybellwaves:
It may not matter but i've seen lowercase true used elsewhere in the docs


beckermr:
It doesn't for yaml but it is more formally correct. One place that does matter is None versus null. Only the second gets parsed to a python None correctly.

PR #2436, opened 2025-02-01
Shorten sentence about `defaults`

jakirkham:
PR Checklist:

### DIFF
--- a/docs/user/introduction.md
+++ b/docs/user/introduction.md
@@ -22,7 +22,7 @@ You can [search](https://anaconda.org/) for packages online. Look out for packag
 
 ## Why conda-forge?
 
-The packaging team from [Anaconda, Inc.](https://anaconda.org/), packages a multitude of packages and provides them to all users free of charge in their `defaults` channel.
+The packaging team from [Anaconda, Inc.](https://anaconda.org/) provides a multitude of packages on their `defaults` channel.
 
 But what if a package you are looking for is not in the `defaults` channel?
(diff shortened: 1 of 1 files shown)


docs/user/introduction.md
```diff
@@ -22,7 +22,7 @@ You can [search](https://anaconda.org/) for packages online. Look out for packag
 
 ## Why conda-forge?
 
-The packaging team from [Anaconda, Inc.](https://anaconda.org/), packages a multitude of packages and provides them to all users free of charge in their `defaults` channel.
+The packaging team from [Anaconda, Inc.](https://anaconda.org/) provides a multitude of packages on their `defaults` channel.
```
jakirkham, line 25: This was likely written a long time ago. Given ToS, this really shouldn't say free of change to all users

However don't want to get into all of that in an intro. So just shortened this instead

Honestly this whole section may benefit from a revamp. In interim this seems like a small reasonable change

beckermr (approved):

beckermr:
We need to fix the dockerhub links in another PR. :/

jakirkham:
Thanks Matt! 🙏

Agreed. Yeah there are plenty of opportunities here to improve the docs

Think a lot of this was written in the mid to late 2010s. Would probably write different things now

For example we are explaining why conda-forge should exist from a standpoint of its creation. Well it does exist and it has so for a while

So maybe we would do better to talk about the communities we are serving

This is just one example 😉

PR #2493, opened 2025-04-02
sort src/core.csv by github handle

h-vetinari:
I was looking at https://github.com/conda-forge/conda-forge.github.io/blob/main/src/core.csv today and noticed that the table isn't sorted. It used to be in the past (c.f. https://github.com/conda-forge/conda-forge.github.io/commit/b7be54cc4f8dab0b0e93e55219cd7b5954390404), so let's sort it.


hmaarrfk (approved):

jaimergp:
We should add a pre-commit hook for this. Something like https://github.com/pre-commit/pre-commit-hooks#file-contents-sorter but skipping the headers.

PR #2539, opened 2025-06-11
add announcement for moving to VS2022

h-vetinari:
As discussed in the core call two weeks ago
https://github.com/conda-forge/conda-forge.github.io/blob/39c236329d7b221165481b0e39ba011da04271e2/community/minutes/2025-05-28.md?plain=1#L53-L59

Now that CUDA 11.8 is dropped, there's no blocker anymore.

### DIFF
--- a/news/2025-06-11-moving-to-vs2022.md
+++ b/news/2025-06-11-moving-to-vs2022.md
@@ -1,0 +1,15 @@
+# Moving to Visual Studio 2022 as default windows compiler
+
+Microsoft's Visual Studio (VS) 2019 compiler has reached its
+[end of life](https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2019)
+over a year ago. In the meantime, several projects have moved on and
+fail to compile with VS2019.
+
+We are planning to update our default compilers on windows to the (fully compatible)
+successor VS2022 in one week from now.
+
+This will not affect you as a general user of conda-forge packages on windows;
+the only potential impact is that if you are compiling locally with VS2019 against
+artefacts produced by conda-forge, you might be required to upgrade.
+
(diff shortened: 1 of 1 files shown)


news/2025-06-11-moving-to-vs2022.md
```diff
@@ -0,0 +1,15 @@
+# Moving to Visual Studio 2022 as default windows compiler
+
+Microsoft's Visual Studio (VS) 2019 compiler has reached its
+[end of life](https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2019)
+over a year ago. In the meantime, several projects have moved on and
+fail to compile with VS2019.
+
+We are planning to update our default compilers on windows to the (fully compatible)
+successor VS2022 in one week from now.
```
hmaarrfk, line 9: is the goal to change "one wekk from from now" to a specific date the second before you press merge?

h-vetinari: Not really. I like to keep a bit of flexibility, because we're rarely exactly punctual with anything we announce 😅 

But if people prefer exact dates (or even times), fine by me.

hmaarrfk: more for posterity, where "next week" no longer has the same context. but fine by me either way.

h-vetinari: If you check https://conda-forge.org/news/, every item has a date (from the file name AFAICT). So "one week" doesn't move around after merging, but it gives us ~1-2 extra days to merge

hmaarrfk: perfect thanks!

PR #2652, opened 2025-11-12
Prune voda from the dev support list

ericdill:
i'm pretty sure that no one still at Votron Data engages with conda-forge any more? The original list when we had that yaml file was myself, @kkraus14 , and @mariusvniekerk . All of whom have moved on. 

Also, is this the right file to just edit directly? used to be a yaml file but the website code seems to have changed since the last time I looked at this


jaimergp:
That's still the case, yep!

jakirkham (approved):

PR #2668, opened 2025-11-25
Add status passing/failing/unknown/... in migration table

Carreau:
This adds a CI status column to https://conda-forge.org/status/migration/?name=python314t, (the value is already present in the data and the graph), and I mostly care about failing ones to help fixing them.

While at it, makes the column sortable because it's nice.
Screenshot 2025-11-25 at 13 49 39

PR #2675, opened 2025-12-01
Share table sorting component & update sortable keys

Carreau:
on top of #2668

-- 
- This extract the sorting tables into their own components.
- Use that also in /status/migrations/?name=....

Therefore it adds logic to also sort by CI status and updated_at,
as ci status is an enum, we now track the previous column we sort with
so that you can sort by status (and break tie by updated_at), or any
other tie break with the second to last sorted column.

CI status and updated_at that are null are always sorted last.

I guess the sorting table coudl take a mapping from field name to a
comparison function, but it would be starting to be a bit of a heavy
handed approach I think.

### DIFF
--- a/src/components/StatusDashboard/current_migrations.jsx
+++ b/src/components/StatusDashboard/current_migrations.jsx
@@ -4,6 +4,7 @@ import React, { useEffect, useState } from "react";
 import { measureProgress } from "@site/src/pages/status/migration";
 import styles from "./styles.module.css";
 import Link from "@docusaurus/Link";
+import { SortableHeader } from "@site/src/components/SortableTable";
 
 const COLLAPSED_KEY = "migration-collapsed";
 const SORT_KEY = "migration-sort";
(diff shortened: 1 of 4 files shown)


src/components/StatusDashboard/current_migrations.jsx
```diff
@@ -230,20 +207,81 @@ function TableContent({ collapsed, name, resort, rows, select, sort, fetched })
...
     case "status":
-      return order === "ascending"
-        ? (a, b) => a.progress.percentage - b.progress.percentage
-        : (a, b) => b.progress.percentage - a.progress.percentage;
+      return (a, b) => {
+        const result = order === "ascending"
+          ? a.progress.percentage - b.progress.percentage
+          : b.progress.percentage - a.progress.percentage;
+        return applySecondarySort(result, a, b);
+      };
```
jaimergp, line 233: In the previous logic, we were returning fat-arrow functions, but in this one we are returning the sorted result, and checking whether the order is ascending or descending in each case. This may result into a slower sort, but maybe not really perceptible UI-wise. I certainly didn't notice it with the amount of data we are displaying.

jaimergp (approved):
LGTM. I have some mild concerns about performance, but they are only theoretical so far. The code is readable and can be iterated upon if necessary later. Thanks @Carreau! This is an excellent UX improvement for the migrations table!

PR #2686, opened 2025-12-08
Diátaxis: cross-compilation

mgorny:
PR Checklist:

-----

This is primarily moving the current docs from KB to a new how-to, and attempting to modernize/clean them up.

I think at least some of the Python bits belong in "understanding conda-forge" — though I suppose to integrate them we'd need to write some more general explanation of how cross works. Not sure how much detail we want to delve into.

Fixes Quansight-Labs/conda-ecosystem-sta-mgmt#60

### DIFF
--- a/docs/maintainer/understanding_conda_forge/cross-compilation.md
+++ b/docs/maintainer/understanding_conda_forge/cross-compilation.md
@@ -128,9 +128,8 @@ about the system it runs on. However, if the executable does not query system in
 build system and the host system are sufficiently similar for the queries run, it's advisable to
 build the executable for the build platform, and for two reasons:
 
-1. Userspace emulation is not available on all platforms. At the moment, it is limited to Linux
-   targets, so for example emulation cannot be used while cross-compiling from `osx-64` to
-   `osx-arm64`.
+1. Userspace emulation is not available on all platforms. For example, emulation cannot be used
+   while cross-compiling from `osx-64` to `osx-arm64`.
 2. Enabling userspace emulation requires superuser privileges, which may make it harder for users to
    build the package locally, in particular to build the build process.
 


docs/how-to/advanced/cross-compilation.mdx
```diff
@@ -0,0 +1,417 @@
...
+
+Note that some resources may use the term "host" to refer to the build machine, and the term
+"target" to refer to the host machine. This is usually the case if they are not referring to
+building compilers.
+
+:::
+
+<a id="cross-compilation-howto"></a>
+
+<a id="how-to-enable-cross-compilation"></a>
```
jaimergp: We can remove these now. They were added in the KB for backwards compatibility reasons, but not needed in the new page.

docs/how-to/advanced/cross-compilation.mdx
```diff
@@ -0,0 +1,417 @@
...
+- `CROSSCOMPILING_EMULATOR`: Path to the `qemu` binary for the host platform. Useful for running
+  tests when cross-compiling.
+
+This is all supported by two main conda-build features introduced in version 3:
+
+- How [requirements metadata](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#requirements-section)
+  is expressed in `meta.yaml`, which distinguishes between `build` and `host` platforms.
+- The `compiler()` Jinja function and underlying [conventions for the compiler packages](https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html).
+
+<a id="placing-requirements-in-build-or-host"></a>
```
jaimergp:
```suggestion
```

docs/how-to/advanced/cross-compilation.mdx
```diff
@@ -0,0 +1,417 @@
...
+The rule of the thumb is:
+
+- If it needs to run during the build, it goes in `build`.
+- If it needs to be available on the target host, it goes in `host`.
+- If both conditions are true, it belongs in both.
+
+However, there are some exceptions to this rule; most notably Python cross-compilation
+([see below](#python-cross-compilation)).
+
+<a id="cross-compilation-examples"></a>
```
jaimergp:
```suggestion
```

docs/how-to/advanced/cross-compilation.mdx
mgorny: Done.

jaimergp:
Let me take a look locally, I think we can pass two children without component tags.

docs/how-to/advanced/cross-compilation.mdx
```diff
@@ -0,0 +1,382 @@
...
+:::note
+
+Some cross-compilation documentation might also distinguish between a third type of platform, the
+target platform. This is used primarily when building cross-compilers, and indicates the platform
+for which the built package will generate code for. For the purposes of this documentation, we'll
+consider this to be irrelevant and the target platform to be the same as the host.
+
+Note that some resources may use the term "host" to refer to the build platform, and the term
+"target" to refer to the host platform. This is usually the case if they are not referring to
+building compilers.
```
isuruf:
```suggestion
"target" to refer to the host platform. This convention is notably used by cmake, but we will not use this convention in this document.
```

docs/how-to/advanced/cross-compilation.mdx
```diff
@@ -0,0 +1,382 @@
...
+- How [requirements metadata](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#requirements-section)
+  is expressed in `meta.yaml`, which distinguishes between `build` and `host` platforms.
+- The `compiler()` Jinja function and underlying [conventions for the compiler packages](https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html).
+
+## Placing requirements in build or host
+
+The rule of the thumb is:
+
+- If it needs to run during the build, it goes in `build`.
+- If it needs to be available on the target host, it goes in `host`.
```
isuruf:
```suggestion
- If it needs to be available during the build, is specific to the `host` machine, but is not run, it goes in `host`. (Eg: headers, libraries, etc)
```

docs/how-to/advanced/cross-compilation.mdx
```diff
@@ -0,0 +1,382 @@
...
+- If both conditions are true, it belongs in both.
+
+However, there are some exceptions to this rule; most notably [Python cross-compilation](#details-about-cross-compiled-python-packages).
```
isuruf: Can you also note that `python` needs to be in `host` even if not cross-compiled because it hasn't historically supported cross-compilation?

mgorny: To be honest, I don't think that's the right place to say that. I mean, people who need to depend on Python while not cross-compiling aren't going to exactly look for where to place it in docs about cross-compiling.

mgorny: Actually, I think I've found a good way to mention it as a note, will do that in a minute.

docs/how-to/advanced/cross-compilation.mdx
mgorny: To be honest, I don't like this. It's not really "rule of thumb" if it effectively repeats the definition of what `host` requirements are.

mgorny:
Okay, I've rebased it and let's review it before I overdo it. I'm open to ideas what else to describe / change.

docs/how-to/advanced/cross-compilation.mdx
```diff
@@ -32,40 +32,39 @@ building compilers.
...
+
+In v0 recipes, `target_platform` is used in place of `host_platform`.
+
 :::note
 
-The `build_platform` and `target_platform` variables are exposed as environment variables in
-the build scripts (e.g. `$build_platform`), and also as Jinja variables in the `meta.yaml`
-selectors (e.g. `# [build_platform != target_platform]`).
+Many existing v1 recipes are using `target_platform` instead of `host_platform`. This works because
+target platform is almost always the same as host platform, though it is technically incorrect.
```
h-vetinari: I think it'd be worth mentioning here that this is really a translation artefact from v0 -> v1

docs/maintainer/understanding_conda_forge/cross-compilation.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. A cross-toolchain, i.e. compiler, linker and other tools that run on the build platform and are
+   capable of producing binaries for the host platform. In some cases (such as GCC) a dedicated
+   cross-compiler is built for every target, whereas in others, such as Clang, a single compiler is
+   capable of producing binaries for multiple targets. The prefix containing these executables
+   is designated by the `${BUILD_PREFIX}` variable, and the tools themselves are frequently prefixed
+   by the host triplet.
+2. The sysroot built for the host platform, including all the libraries and header files needed by
+   the built package. This sysroot is designated by the `${PREFIX}` variable. In conda-forge, it is
+   also used as the install location for the built package files.
```
h-vetinari: > This sysroot is designated by the `${PREFIX}` variable.

Can you explain what you mean here? The actual sysroot is also under a target-specific folder, e.g.
```
$PREFIX/x86_64-conda-linux-gnu/sysroot/...
```

docs/maintainer/understanding_conda_forge/cross-compilation.md
```diff
@@ -0,0 +1,63 @@
...
+characteristics of the host platform need to be used from `${PREFIX}`, and specified as `host`
+dependencies.
+
+In conda-forge, the build setup and compiler activation scripts attempt to configure the build
+process for correct locations, but in some cases (e.g. when dealing with custom `*-config` tools),
+additional effort may be required from feedstock maintainers to point the build system at the
+correct locations.
+
+While performing cross-builds, the build process cannot run executables built by the cross-compiler,
+for host platform. This has various implications, such as:
```
h-vetinari: "for host platform" sounds weird.

docs/maintainer/understanding_conda_forge/cross-compilation.md
```diff
@@ -0,0 +1,63 @@
...
+
+- Programs are unable to run their test suites. Testing is often skipped when cross-compiling, or
+  emulators are used to run them.
```
h-vetinari: I think this document could add a brief explanation of emulators, since it's a key part of our cross-compilation story on linux (e.g. no work is necessary to use an emulator for testing, it all works out of the box)

docs/maintainer/understanding_conda_forge/cross-compilation.md
```diff
@@ -98,3 +98,26 @@ for host platform. This has various implications, such as:
...
+- A modified `crossenv` installation in `$BUILD_PREFIX/venv`, mimicking the architecture of
+  `$PREFIX`.
+- Forwarder binaries in `$BUILD_PREFIX/bin` that point to the `crossenv` installation.
+- Symlinks that expose the `$BUILD_PREFIX` site-packages in the `crossenv` installation, which
+  is also included in `$PYTHONPATH`.
+- A copy of all `$PREFIX` site-packages to `$BUILD_PREFIX` (except the compiled libraries).
+
+All in all, this results in a setup where `conda-build` can run a `$BUILD_PREFIX`-architecture
+`python` interpreter that can see the packages in `$PREFIX` (with the compiled bits provided by
+their corresponding counterparts in `$BUILD_PREFIX`) and sufficiently mimic that target
```
h-vetinari, line 178: missing period

docs/maintainer/understanding_conda_forge/cross-compilation.md
```diff
@@ -0,0 +1,150 @@
...
+floating-point unit.
+
+## Platform types
+
+When discussing toolchains, three types of platform are referenced:
+
+- The build platform, referring to the platform on which the cross-toolchain is being run.
+- The host platform, referring to the platform for which the binaries are being produced.
+- The target platform, relevant to building a cross-toolchain itself, and referring to the platform
+  for which the resulting toolchain will produce binaries.
```
h-vetinari:
```suggestion
- The target platform, which is only relevant if the binaries being produced will _themselves_ target
  some specific architecture. This is generally only needed when building a cross-toolchain, in which
  case it refers to the platform for which the resulting toolchain will produce binaries.
```

docs/maintainer/understanding_conda_forge/cross-compilation.md
h-vetinari: I went commit-by-commit, so this was resolved later already

docs/maintainer/understanding_conda_forge/cross-compilation.md
mgorny: Actually, the sentence got cut. Fixed now.

docs/maintainer/understanding_conda_forge/cross-compilation.md
mgorny: Reworded. I suppose we don't need to name it "host platform" explicitly, and we can focus on the fact that it's a different platform.

docs/maintainer/understanding_conda_forge/cross-compilation.md
mgorny: Well, https://conda-forge.org/docs/maintainer/knowledge_base/#emulated-builds doesn't seem to cover that part, so I guess I need to research a bit — especially that we're mentioning skipping the test phase when cross-compiling.

mgorny: Okay, I've tested a bit and I now see how things are supposed to work. I'll update both docs to make this clearer.

docs/maintainer/understanding_conda_forge/cross-compilation.md
```diff
@@ -109,6 +110,24 @@ since they are built for a different platform. This has various implications, su
...
+`binfmt_misc` permits executing non-native executables directly.
+
+That said, there is no guarantee that an emulator will actually be available. Therefore, it is
+important that the recipe can be built successfully without an emulator. Parts such as tests can
+require an emulator, but they must be guarded accordingly, and run only for native builds or when
+`CROSSCOMPILING_EMULATOR` is not empty.
+
+Furthermore, the emulator introduces an overhead for running binaries, and does not guarantee 100%
+correctness for all cases. Therefore, whenever it makes sense, binaries built for the build platform
+should be preferred. To ensure correct test results, unittests should be compiled for the host
```
isuruf: > Therefore, whenever it makes sense, binaries built for the build platform should be preferred

This statement is confusing. Can you elaborate more?

mgorny: My thinking was that if tests call some external commands, you don't need to replace all of them with host system executables (say, the POSIX shell utilities) — you only need `host` dependencies where using `build` dependency would cause the wrong code to be tested.

h-vetinari: I'd tend to drop that whole paragraph, except perhaps the performance bit. E.g.

```
Note that emulation is substantially slower (roughly 5-6x longer runtime) than cross-compilation.
For this reason alone, cross-compilation is preferable in most cases, despite the above-mentioned limitations.
```

isuruf: That's also confusing. You are conflating full system emulation (i.e. `provider: linux_aarch64: linux_64`) with user-space emulation (i.e. `build_platform: linux_aarch64: linux_64`). This whole section is about latter. Something along the lines of

```
Therefore for linux, you don't need to build executables for `build` platform if this executable needs to be run and can rely on the executable for `host` platform to be run in CI. This is useful if the executable queries some information about the system it runs on. However if the executable does not query system information or if the build system and the host system are sufficiently similar for the queries run, it's advisable to build the executable for `build` platform due to two reasons.
- On systems where there's no emulation (compiling for osx-arm64 from osx-64), there's no other solution.
- With emulation, building the recipe on user systems would require sudo rights to turn on binfmt and makes it harder for users to debug the recipe
```

mgorny: Updated. I've tried to incorporate the feedback from both of you, and since this is the "understanding" part, I've also shortly explained the two kinds of emulation.

docs/maintainer/understanding_conda_forge/cross-compilation.md
```diff
@@ -0,0 +1,180 @@
...
+
+This could eliminate the need to build executables for the build platform, as the executables built
+for the host platform can be run in CI. This is useful if the executable queries some information
+about the system it runs on. However, if the executable does not query system information, or if the
+build system and the host system are sufficiently similar for the queries run, it's advisable to
+build the executable for the build platform, and for two reasons:
+
+1. Userspace emulation is not available on all platforms. At the moment, it is limited to Linux
+   targets, so for example emulation cannot be used while cross-compiling from `osx-64` to
+   `osx-arm64`.
```
isuruf: Technically `osx-arm64` to `osx-64` works with Rosetta2

mgorny: Hmm, so you're suggesting removing "limited to Linux targets"?

mgorny: Done.

isuruf:
Thanks

PR #2696, opened 2025-12-17
Update "contributing packages" for v1 recipes

mgorny:
PR Checklist:

I've basically wanted to provide v1 snippets, but one thing lead to another, and I've started actually rewriting the text surrounding them to be more builder-agnostic, and now we have a full update that covers both v1 and v0 recipes. It's not perfect, but I'm not sure if the aim is to make it perfect. I feel like we're going to split it into tutorials and how-tos in the future.

Part of https://github.com/Quansight-Labs/conda-ecosystem-sta-mgmt/issues/102

### DIFF
--- a/docs/maintainer/adding_pkgs.mdx
+++ b/docs/maintainer/adding_pkgs.mdx
@@ -809,7 +809,7 @@ test:
   requires:
     - pytest
   source_files:
-    - tests
+    - tests/
     - test_pkg_integration.py
   commands:
     - pytest tests/
@@ -826,7 +826,7 @@ tests:
         - pytest
     files:
       source:
-        - tests
+        - tests/
         - test_pkg_integration.py
     script:
       - pytest tests/


docs/maintainer/adding_pkgs.mdx
```diff
@@ -264,16 +266,26 @@ Once these steps are complete, you can continue with the steps in [Step-by-step
...
 
-The `meta.yaml` file in the recipe directory is at the heart of every conda package.
 It defines everything that is required to build and use the package.
 
-`meta.yaml` is in [yaml](https://en.wikipedia.org/wiki/YAML) format, augmented with [Jinja](http://jinja.pocoo.org/) templating.
+Both files use the [yaml](https://en.wikipedia.org/wiki/YAML) format, augmented with [Jinja](http://jinja.pocoo.org/) templating.
 
-A full reference of the structure and fields of `meta.yaml` file can be found in the [Defining metadata (meta.yaml)](https://conda.io/projects/conda-build/en/stable/resources/define-metadata.html) section in the conda-build documentation.
+The full reference of the formats can be found in the following sources:
+- `recipe.yaml`: [rattler-build: the recipe spec](https://rattler-build.prefix.dev/latest/reference/recipe_file/)
```
isuruf: It's important to note the distinction between v1 format and rattler-build. This sentence makes the user think they are tightly coupled even though they are not.

mgorny: Do you have a specific suggestion on how to do that? I think I get your point, but I'm not really saying that they are coupled, or even saying much about v0/v1 at this point. I could change the link descriptions to avoid saying "conda-build" and "rattler-build", but that feels quite artificial, given that these are precisely the docs we're linking to.

docs/maintainer/adding_pkgs.mdx
```diff
@@ -41,7 +44,37 @@ The staging process i.e adding a package's recipe has three steps:
...
+[rattler-build generate-recipe command](https://rattler-build.prefix.dev/latest/recipe_generation/).
+At the time of writing, the tool supports generating recipes for:
+
+- Python packages from PyPI
+- R packages from CRAN
+- Perl packages from CPAN
+- Lua packages from LuaRocks
+
+For a full list, consult the output of `rattler-build generate-recipe`. Note that the generated
+recipe needs to be reviewed and modified as necessary.
```
mgorny: Isuru pointed out that graykull is a better choice for v1 recipes as well. I'll update this next week.

docs/maintainer/adding_pkgs.mdx
h-vetinari: Well, the spec is defined by CEP 13 & 14. Work is [progressing](https://github.com/conda/conda-build/pull/5880) to support v1 in conda-build (even if it uses rattler under the hood), for example. I think the prefix docs are a great reference, but they're not the spec.
```suggestion
- `recipe.yaml`: improved recipe format as specified by [CEP 13](https://github.com/conda/ceps/blob/main/cep-0013.md) & [CEP 14](https://github.com/conda/ceps/blob/main/cep-0014.md). As of early 2026, it is only implemented by rattler-build, which also provides a more readable [overview](https://rattler-build.prefix.dev/latest/reference/recipe_file).
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -337,33 +352,65 @@ curl -sL https://github.com/username/reponame/archive/vX.X.X.tar.gz | openssl sh
...
+      - win
+      - osx-arm64
   ```
+  </RecipeTabs>
 - for specific python versions:
+  <RecipeTabs>
+
   ```yaml
   build:
       skip: true  # [py<35]
```
h-vetinari: Let's use something more current here, like `<311`

docs/maintainer/adding_pkgs.mdx
```diff
@@ -337,33 +352,65 @@ curl -sL https://github.com/username/reponame/archive/vX.X.X.tar.gz | openssl sh
...
 <a id="optional-bld-bat-and-or-build-sh"></a>
 
-#### Optional: `bld.bat` and/or `build.sh`
+#### Optional: build scripts
 
-In many cases, `bld.bat` and/or `build.sh` files are not required.
+In many cases, explicit build scripts are not required.
 Pure Python packages almost never need them.
 
+The build scripts are named:
```
h-vetinari:
```suggestion
The default names for the build scripts are
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -515,25 +575,48 @@ requirements:
...
 
 ```yaml
 package:
-name: awesome-db
+  name: awesome-db
 
 requirements:
   # [...]
   run_constrained:
     - amazing-db ==9999999999
```
h-vetinari: We've moved on from the 999... (which could theoretically be satisfied) towards
```suggestion
    - amazing-db <0.0a0
```
which cannot.

docs/maintainer/adding_pkgs.mdx
```diff
@@ -563,44 +646,75 @@ However, this is no reason for the recipe to not have tests. At the very least,
 we want to verify that the package has installed the desired files in the desired
 locations. This is called existence testing.
 
-Existence testing can be accomplished in the `meta.yaml` file in the
-`test/commands` block.
+Existence testing can be accomplished using `script` tests in v1 recipes, or `commands` subsection
+of the `test` section of v0 recipe.
```
h-vetinari: This completely misses the `package_content.files:` directive in v1 recipes, see https://rattler-build.prefix.dev/latest/testing/#testing-package-contents. In particular `test -f ...` and `if not exist ...` should not be necessary anymore for the vast majority of v1 recipes

docs/maintainer/adding_pkgs.md
```diff
@@ -614,23 +728,6 @@ that do contain executable code (e.g. `package_name.core`).
...
-<a id="id8"></a>
-
-##### pip check
-
-For PyPI packages, we strongly recommend including `pip check` as part of the `test.commands` section:
-
-```yaml
-test:
-  commands:
-    - pip check
```
h-vetinari, line 628: This should not be removed; v1 even has a specific command for this (it's on by default, that needs to be described)
```
  # python specific tests
  - python:
      # this test section tries to import the python modules and errors if it can't
      imports:
        - mypkg
      pip_check: true
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -660,15 +757,32 @@ to simply use the package. This is usually a test-running framework, such as
 nose or pytest. You can ensure that it is included by adding it to requirements
```
h-vetinari: let's remove mentions of nose where we find them

docs/maintainer/adding_pkgs.mdx
```diff
@@ -689,10 +806,28 @@ test:
     - tests
     - test_pkg_integration.py
   commands:
-    - pytest tests test_pkg_integration.py
+    - python -m pytest tests test_pkg_integration.py
```
h-vetinari: what's the case for `python -m`? In a correct environment containing python & pytest, the following is unambiguous. Also I'd prefer to indicate folders with a trailing slash, and to break this into two lines
```suggestion
    - pytest -v tests/
    - pytest -v test_pkg_integration.py
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -689,10 +806,28 @@ test:
...
+        - package_name
+  - requirements:
+      run:
+        - pytest
+    files:
+      source:
+        - tests
+        - test_pkg_integration.py
+    script:
+      - python -m pytest tests test_pkg_integration.py
```
h-vetinari: same here

docs/maintainer/adding_pkgs.mdx
```diff
@@ -718,27 +864,50 @@ will be run at test time. This allows an arbitrarily complicated test script.
...
 
+<RecipeTabs>
+
 ```yaml
 test:
   requires:
     - pytest
   commands:
-    - pytest --pyargs package_name
+    - python -m pytest --pyargs package_name
```
h-vetinari: This is now redundant with the very first example you expanded. I'd prefer to keep this with the explanation that the tests need to be _installed_  for this to work (and perhaps how that's not necessarily always the best idea, especially if the `tests/` folder is big)

docs/maintainer/adding_pkgs.mdx
```diff
@@ -798,6 +979,23 @@ test:
     - "\"%R%\" -e \"testthat::test_file('tests/testthat.R', stop_on_failure=TRUE)\""  # [win]
```
h-vetinari: In fairness to v0, if we're calling bare `R` in the v1 recipe, we can do the same for v0

docs/maintainer/adding_pkgs.mdx
```diff
@@ -1016,7 +1224,6 @@ Then, include the tool as a build time dependency.
 ```yaml
 requirements:
   build:
-    - {{ compiler('go') }}
```
h-vetinari, line 1019: why are you removing this? `go build` without a go compiler will not work

docs/maintainer/adding_pkgs.mdx
```diff
@@ -1161,8 +1387,23 @@ Jinja expressions serve following purposes in the meta.yaml:
...
 
+  ```yaml
+  requirements:
+    build:
+      - ${{ compiler('c') }}
+      - ${{ compiler('cxx') }}
+      - if: linux
+        then:
+          - ${{ cdt('xorg-x11-proto-devel') }}
+          - ${{ cdt('libx11-devel') }}
```
h-vetinari: Please add stdlib and remove the CDTs (also for v0)

docs/maintainer/adding_pkgs.mdx
```diff
@@ -1175,4 +1416,18 @@ Jinja expressions serve following purposes in the meta.yaml:
       - python
   ```
 
-For more information please refer to the [Templating with Jinja](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#templating-with-jinja) section in the conda-build docs.
+  ```yaml
+  requirements:
+    build:
+      - ${{ compiler('c') }}
```
h-vetinari, line 1448: Please add stdlib; also for v0
```suggestion
      - ${{ stdlib('c') }}
      - ${{ compiler('c') }}
```

h-vetinari (changes requested):
Thanks for the work on this! Did a first pass, a few things need sharpening

docs/maintainer/adding_pkgs.mdx
mgorny: Nice. I didn't know about that!

docs/maintainer/adding_pkgs.mdx
mgorny: I was wondering why we're going through a variable there but not in v1.

docs/maintainer/adding_pkgs.md
mgorny: It's merged into the block above, since you can't do `pip_check` without `imports:` in v1 recipe (https://github.com/prefix-dev/rattler-build/issues/2063).

docs/maintainer/adding_pkgs.mdx
mgorny: Hmm, I don't recall why I did that. Perhaps a side effect of some back and forth.

I suppose nothing wrong with splitting but that's IMO suboptimal — you end up loading `conftest.py` twice.

docs/maintainer/adding_pkgs.mdx
mgorny: Ah, my bad. I was assuming we just want to list the package providing `go-licenses`, but indeed `go build` is present in the snippet above.

docs/maintainer/adding_pkgs.mdx
h-vetinari: > you end up loading `conftest.py` twice.

...and? Do you mean to say that this has a noticeable performance impact? They'd be completely independent calls, there are more things duplicated (interpreter startup, etc.). I think that's completely negligible. 

I prefer this because it's simpler. It says "look, I can test different things, first the test suite, then an integration test", without having to understand the details of pytests test collection. 

Ultimately not a big deal, but across those docs I'd really aim for as-simple-as-possible within the context of the point being made

docs/maintainer/adding_pkgs.mdx
mgorny: I'm confused. It precisely says above the snippet "If the tests are installed with the package". Or are you suggesting we add a mention that one may want to actually remove the tests from the installed package?

h-vetinari: > This is now redundant with the very first example you expanded.

The "very first example" here refers to lines 760-785 (counting in green), where you've now added the same `pytest --pyargs package_name`, only that it's not explained what this does, much less "arrived at" as a less common variant.

docs/maintainer/adding_pkgs.mdx
mgorny: Oh, I guess I removed it, because it saved us from the v0/v1 split.

docs/maintainer/adding_pkgs.mdx
mgorny: Sorry, I mean: I was wondering why v1 format removed the variable that was used in the v0 format. Do you happen to know if it used to be actually required at some point?

h-vetinari: I don't know all the history, but lots of recipes were initially using `$PYTHON`, presumably as a safety to ensure that we're really picking up the right python from the environment, and not something else randomly flying around. I assume the situation for R was similar. 

Things have gotten much better w.r.t. "randomly picking stuff up", so nowadays, recipes can just use `python`. For extra brownie points, could you find some random R-recipe still using `$R` / `%R%` (org-wide regex search is super useful for that), replace the variable and see if it works?

My original point was that if it works for v1 nowadays, there's no reason why it shouldn't work for v0 too, and so we should reflect that also in the v0 portion. The shell invocation logic is not so fundamentally different (at least, that's my expectation).

docs/maintainer/adding_pkgs.mdx
mgorny: Oh. I've added "some" invocation there because it felt wrong to be listing a test block with a dependency but without any command, and I've used `--pyargs` there because I didn't want to introduce `source_files:` at that point — i.e. that example focuses on dependencies, whereas this one focuses on pytest invocation.

Hmm, how about I change the first example to list some other test dependency and change invocation to plain `python -m unittest`?

docs/maintainer/adding_pkgs.mdx
mgorny: Tested linux locally, and filed https://github.com/conda-forge/r-catboost-feedstock/pull/88 to test Windows.

docs/maintainer/adding_pkgs.mdx
```diff
@@ -657,18 +754,35 @@ not be run straight from an installed package.
...
+
 ```yaml
 test:
   imports:
     - package_name
-...
   requires:
     - pytest
+  commands:
+    - pytest --pyargs package_name
```
h-vetinari: I think the easiest solution might be
```suggestion
    # will be explained below in more detail
    - pytest --pyargs package_name
```

Or simply
```suggestion
    - ... # details below
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -515,25 +575,48 @@ requirements:
...
+```
+
+```yaml
+package:
+  name: awesome-db
+
+requirements:
+  # [...]
+  run_constraints:
     - amazing-db ==9999999999
```
h-vetinari: I would have thought this self-evident, but unless stated otherwise, my reviews for one side of v0/v1 divide generally apply to the other as well.

(replace with `<0.0a0`)

docs/maintainer/adding_pkgs.mdx
h-vetinari, line 1413: Missing stdlib, remove CDTs

h-vetinari, line 1435: missing stdlib (etc.)

docs/maintainer/adding_pkgs.md
h-vetinari: OK, I didn't notice this. Thanks

docs/maintainer/adding_pkgs.mdx
h-vetinari: Responded in a separate [comment](https://github.com/conda-forge/conda-forge.github.io/pull/2696#discussion_r2675445192)

docs/maintainer/adding_pkgs.mdx
mgorny: Ah, sorry, clicked through GH and missed that.

docs/maintainer/adding_pkgs.mdx
```diff
@@ -85,24 +88,26 @@ In case you are building your first recipe using conda-forge, a step-by-step ins
 1. Ensure your source code can be downloaded as a single file. Source code
    should be downloadable as an archive (.tar.gz, .zip, .tar.bz2, .tar.xz)
    or tagged on GitHub, to ensure that it can be verified. (For further
-   detail, see [Build from tarballs, not repos](#tarballs-no-repos)).
+   detail, see [Build from tarballs, not repos](#build-from-tarballs-not-repos)).
 2. Fork and clone the [staged-recipes](https://github.com/conda-forge/staged-recipes)
    repository from GitHub.
 3. Checkout a new branch from the staged-recipes `main` branch.
-4. Through CLI, cd inside the ‘staged-recipes/recipes' directory.
+4. Through CLI, cd inside the `staged-recipes/recipes` directory.
```
jaimergp:
```suggestion
4. In the same terminal, go to the `staged-recipes/recipes` directory.
```

A bit pedantic but I guess this is a bit more correct. Using `cd` as a verb is weird.

jaimergp: Alternatively:

```suggestion
4. In the terminal, `cd` inside the `staged-recipes/recipes` directory.
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -85,24 +88,26 @@ In case you are building your first recipe using conda-forge, a step-by-step ins
...
+6. [Generate the recipe](#generating-the-recipe) or copy the example `recipe.yaml` (or `meta.yaml`)
+   into this folder.
+   All the changes in the following steps will happen in the COPIED `recipe.yaml` or `meta.yaml`
+   (i.e., `...staged-recipes/recipes/<name-of-package>/recipe.yaml`).
    Please leave the example directory unchanged!
-7. Modify the copied recipe (meta.yml) as needed. To see how to modify meta.yaml, take a look at
-   [The recipe meta.yaml](#id4).
-8. Generate the SHA256 key for your source code archive, as described in the
+7. Modify the copied recipe file as needed. To see how to modify it, take a look at
+   [The recipe (recipe.yaml or meta.yaml)](#the-recipe-recipeyaml-or-metayaml).
```
jaimergp:
```suggestion
   [The recipe (recipe.yaml or meta.yaml)](#the-recipe-recipeyaml-or-metayaml) section.
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -85,24 +88,26 @@ In case you are building your first recipe using conda-forge, a step-by-step ins
...
+8. Generate the SHA256 hash for your source code archive, as described in the
    example recipe using the `openssl` tool. As an alternative, you can also
-   go to the package description on [PyPi](https://pypi.org) from which you
+   go to the package description on [PyPI](https://pypi.org) from which you
    can directly copy the SHA256.
 9. Be sure to fill in the `test` section. The simplest test will simply
    test that the module can be imported, as described in the example.
```
jaimergp:
```suggestion
9. Be sure to fill in the `test` section. The simplest test will simply
   test that the top-level package or module can be imported, as described in the example. 
   Some projects have an empty top-level package; in those cases, make sure to select also
   some other modules that do contain runnable code.
```

A pet peeve of mine; importing empty `__init__` which do not really do much.

docs/maintainer/adding_pkgs.mdx
```diff
@@ -118,11 +123,11 @@ curl -sL https://github.com/username/reponame/archive/vX.X.X.tar.gz | openssl sh
 
 #### Checklist
 
-- Ensure that the license and license family descriptors (optional) have the right case and that the license is correct. Note that case sensitive inputs are required (e.g. Apache-2.0 rather than APACHE 2.0). Using SPDX identifiers for license field is recommended. (see [SPDX Identifiers and Expressions](#spdx))
+- Ensure that the license has the right case and that the license is correct. Note that case sensitive inputs are required (e.g. `Apache-2.0` rather than `APACHE 2.0`). Using SPDX identifiers for license field is recommended. (see [SPDX Identifiers and Expressions](#spdx-identifiers-and-expressions))
 - Ensure that you have included a license file if your license requires one – most do. (see [here](https://github.com/conda-forge/staged-recipes/blob/a504af81c05491bf7b0b018b2fa1efe64767985c/recipes/example/meta.yaml#L52-L55))
```
jaimergp: May want to update this link to `recipe.yaml` _and_ `meta.yaml` files, I guess?

docs/maintainer/adding_pkgs.mdx
```diff
@@ -258,22 +263,32 @@ staged-recipes directory.
...
+and is named either:
+
+- `recipe.yaml` for the newer v1 recipes
+- `meta.yaml` for the older v0 recipes
 
-The `meta.yaml` file in the recipe directory is at the heart of every conda package.
 It defines everything that is required to build and use the package.
 
-`meta.yaml` is in [yaml](https://en.wikipedia.org/wiki/YAML) format, augmented with [Jinja](http://jinja.pocoo.org/) templating.
+Both files use the [yaml](https://en.wikipedia.org/wiki/YAML) format, augmented with [Jinja](http://jinja.pocoo.org/) templating.
```
jaimergp:
```suggestion
Both files use [YAML](https://en.wikipedia.org/wiki/YAML) syntax, augmented with [Jinja](http://jinja.pocoo.org/) templating.
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -327,7 +342,7 @@ curl -sL https://github.com/username/reponame/archive/vX.X.X.tar.gz | openssl sh
 
 #### Downloading extra sources and data files
 
-`conda-build 3` supports multiple sources per recipe. Examples are available [in the conda-build docs](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#source-from-multiple-sources).
+`conda-build 3` and `rattler-build` support multiple sources per recipe. Examples are available in the [rattler-build](https://rattler-build.prefix.dev/latest/reference/recipe_file/#source-section) and [conda-build docs](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#source-from-multiple-sources).
```
jaimergp:
```suggestion
`conda-build` (v3 and above) and `rattler-build` support multiple sources per recipe. Examples are available in the [rattler-build](https://rattler-build.prefix.dev/latest/reference/recipe_file/#source-section) and [conda-build docs](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#source-from-multiple-sources).
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -337,33 +352,65 @@ curl -sL https://github.com/username/reponame/archive/vX.X.X.tar.gz | openssl sh
...
 
-Use the `skip` key in the `build` section along with a selector:
+Use the `skip` key in the `build` section. In v1 recipes, it accepts any of:
+
+- a boolean value, such as `true` to skip the build unconditionally;
+- a selector expression to skip the build when it evaluates to true;
+- a list of selector expressions, in which case the build is skipped if any of the expressions
+  evaluates to true.
+
+In v0 recipes, only a boolean value is directly accepted, and it needs to be combined with
```
jaimergp, line 364: I guess we could mention that sometimes you can also have multiple `skip: true` lines with different selectors instead of a long chain of top-level `or` clauses.

docs/maintainer/adding_pkgs.mdx
```diff
@@ -563,44 +647,76 @@ However, this is no reason for the recipe to not have tests. At the very least,
...
+tests:
+  - python:
+      imports:
+        - package_name
+```
+
+</RecipeTabs>
+
+Note that `package_name` is the name imported by Python;
 not necessarily the name of the conda package (they are sometimes different).
```
jaimergp:
```suggestion
not necessarily the name of the conda package (they are sometimes different), or even the wheel name.
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -657,18 +756,35 @@ not be run straight from an installed package.
 
 Sometimes there are packages required to run the tests that are not required
 to simply use the package. This is usually a test-running framework, such as
-nose or pytest. You can ensure that it is included by adding it to requirements
+pytest. You can ensure that it is included by adding it to requirements
```
jaimergp:
```suggestion
`pytest`. You can ensure that it is included by adding it to requirements
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -701,13 +840,24 @@ The `source_files` section works for files and directories.
...
+```yaml
+tests:
+  ...
+  - script:
+      - python -c "import package_name; package_name.tests.runall()"
 ```
 
+</RecipeTabs>
+
 Alternatively, you can add a file called `run_test.py` in the recipe that
```
jaimergp: I think the presence of this file overrides any `test` section (at least in `conda-build`), which may be considered a footgun. We should double check whether this is also the case in rattler-build, and add a small warning / note.

docs/maintainer/adding_pkgs.mdx
```diff
@@ -726,19 +878,40 @@ test:
...
+      - pytest --pyargs package_name
+```
+
+</RecipeTabs>
+
 <a id="command-line-utilities"></a>
 
 #### Command Line Utilities
 
 If a python package installs command line utilities, you probably want to test that
```
jaimergp:
```suggestion
If a Python package installs command line utilities, you probably want to test that
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -831,7 +1030,7 @@ present and needed for specific tests.
 The default R build procedure on conda-forge will not include the
 `tests/` directory in the final build. While it is possible to do this
 (via an `--install-tests` flag), it is preferable to use the
-`tests.source_files` in the `meta.yaml` to copy the tests for the
+`files.source` or `source_files` keys to copy the tests for the
```
jaimergp: This is for `recipe.yaml` and `meta.yaml`, respectively, right?


```suggestion
`files.source` or `source_files` keys (for `recipe.yaml` or `meta.yaml`, respectively) to copy the tests for the
```

docs/maintainer/knowledge_base.md
```diff
@@ -2197,7 +2197,7 @@ being investigated or the migration PR got closed accidentally. To migrate a PR
 All conda-forge packages are built by strangers on the internet on public cloud infrastructure from source code you likely have not inspected, so you should not use conda-forge packages if you or your team require a high level of security.
 You are also free to download recipes and rebuild them yourself, if you would like at least that much oversight. However, many people use conda-forge all the time with no issues and here are some things that conda-forge does to help with security in some ways:
 
-1. [Sources](adding_pkgs.md#meta-yaml-source) (where you specify where the package's source code is coming from) can be pulled from GitHub, PyPI, or other sources and sha256 hashes are always used, so moving of tags or uploading of new sdists can not cause automatic package rebuilds.
+1. [Sources](../adding_pkgs#source) (where you specify where the package's source code is coming from) can be pulled from GitHub, PyPI, or other sources and sha256 hashes are always used, so moving of tags or uploading of new sdists can not cause automatic package rebuilds.
```
jaimergp:
```suggestion
1. [Sources](../adding_pkgs#source) (where you specify where the package's source code is coming from) can be pulled from GitHub, PyPI, or other sources and SHA256 hashes are always used, so moving of tags or uploading of new sdists can not cause automatic package rebuilds.
```

docs/user/contributing.mdx
```diff
@@ -13,7 +13,7 @@ conda-forge is a community-driven effort of cross-platform packaging and relies
 
 We encourage you to contribute to conda-forge. You can do so in several ways:
 
-- [Contribute new packages](../maintainer/adding_pkgs.md).
+- [Contribute new packages](../../maintainer/adding_pkgs).
```
jaimergp, line 16: Hm, why is this extra parent needed here but not below? 🤔

docs/maintainer/adding_pkgs.mdx
mgorny: According to [rattler-build docs](https://rattler-build.prefix.dev/v0.51.0/testing/#legacy-tests), they are "still supported for execution", but I have no clue what that means. If I put a `recipe/run_test.py` file, it doesn't get run — with or without `tests:` section.

mgorny: Apparently in v1 recipe, the equivalent is a `script` test with `file:` key. Updated for that.

docs/user/contributing.mdx
mgorny: It's needed when we don't explicitly give the `.mdx` suffix. I figured out avoiding this makes the links more future proof.

docs/maintainer/adding_pkgs.mdx
```diff
@@ -515,26 +579,50 @@ requirements:
...
+  # [...]
+  run_constraints:
+    - amazing-db <0.0a0
 ```
 
-In this example, `awesome-db` cannot be installed together with `amazing-db` as there is no package `amazing-db-9999999999`.
+</RecipeTabs>
+
+In this example, `awesome-db` cannot be installed together with `amazing-db` as there cannot be
+an `amazing-db <0.0a0` package.
```
h-vetinari:
```suggestion
In this example, `awesome-db` cannot be installed together with `amazing-db` as the constraint
`amazing-db <0.0a0` is impossible to satisfy.
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -689,10 +812,30 @@ test:
...
+tests:
+  - python:
+      imports:
+        - package_name
+  - requirements:
+      run:
+        - pytest
+    files:
+      source:
+        - tests
```
h-vetinari: This is very much a minor nit, but just for visual clarity I'd prefer to have folders in `files:` denoted with a trailing slash
```suggestion
      source:
        - tests/
```

docs/maintainer/adding_pkgs.mdx
```diff
@@ -689,10 +812,30 @@ test:
     - tests
```
h-vetinari: same here

jaimergp (approved):

PR #2697, opened 2025-12-19
Add prefix dev to cloud services

afshin:
This PR adds the monitor status for `repo.prefix.dev` to the cloud services section of the status page. I am not 100% confident that `repo.prefix.dev` is the correct monitor to query, if someone can authoritatively comment on this I would appreciate it!

The monitoring API they use is UptimeRobot and the format most of the other services we query use is StatusPage, so we have a small section of logic that normalizes the status of these API responses. I updated this snippet of logic to support UptimeRobot monitor responses.

PR Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -12,7 +12,7 @@ jobs:
     name: deploy
     runs-on: "ubuntu-latest"
     steps:
-      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
       - uses: prefix-dev/setup-pixi@82d477f15f3a381dbcc8adc1206ce643fe110fb7 # v0.9.3
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
       - id: prettier
         types_or: [markdown, mdx, json, yaml, css, javascript]
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.9
+    rev: v0.14.10
     hooks:
       - id: ruff-format
 ci:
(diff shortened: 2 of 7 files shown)


wolfv (approved):

SylvainCorlay:
> When is this useful in the context of conda-forge? AFAIK we don't use any of the APIs mentioned in the status page in conda-forge.

@isuruf I missed your reply. In miniforge, Mamba falls back to `prefix.dev` when `anaconda.org` does not work.

src/constants.js
```diff
@@ -58,6 +58,11 @@ export const urls = {
       link: "https://ci-status.quansight.dev/",
       title: "Open GPU Server",
     },
+    prefix_dev: {
+      api: "https://status.prefix.dev/api/getMonitorDetails/status.conda-forge.org?m=798888560",
+      link: "https://status.prefix.dev",
+      title: "repo.prefix.dev",
```
isuruf:
```suggestion
      title: "conda-forge mirror on prefix.dev",
```

afshin:
Thanks for the suggestion @isuruf!

src/constants.js
```diff
@@ -58,6 +58,11 @@ export const urls = {
       link: "https://ci-status.quansight.dev/",
       title: "Open GPU Server",
     },
+    prefix_dev: {
+      api: "https://status.prefix.dev/api/getMonitorDetails/status.conda-forge.org?m=798888560",
+      link: "https://status.prefix.dev",
+      title: "conda-forge mirror on prefix.dev",
```
jaimergp, line 64: This title breaks the "alphabetic" order in the list. What about:

```suggestion
      title: "Prefix.dev mirror",
```

isuruf: It's less informative. Is conda-forge keeping a prefix.dev mirror? Or the other way around?

jaimergp: This is in the "Cloud services" table, none of those services (anaconda.org, Azure, Quay...) are maintained by conda-forge. I don't know if anyone would understand it as such.

isuruf: Those services are what is used by conda-forge on its feedstocks. This is different.

I don't understand why you'd go for a less informative title just because of a need to please a cosmetic "alphabetic" order.

isuruf:
Thanks

PR #2701, opened 2025-12-23
Start explanations on "compiler concepts"

mgorny:
PR Checklist:

Fixes https://github.com/Quansight-Labs/conda-ecosystem-sta-mgmt/issues/67

Per @jaimergp 's suggestions.

### DIFF
--- a/docs/maintainer/understanding_conda_forge/compilation-concepts.md
+++ b/docs/maintainer/understanding_conda_forge/compilation-concepts.md
@@ -346,3 +346,43 @@ quite complex. However, for our purposes it suffices to list the following varia
 Note that steps 2. and 5. specifically focus on program directories. To account for this,
 conda-forge generally installs `.dll` libraries into program directories such as the `bin` directory
 rather than the `lib` directory used on Unixes.
+
+## Architecture-dependent and architecture-independent packages
+
+Conda-forge packages can be built for specific architectures, or made architecture-independent, also
+known as `noarch`.
+
+When C code is compiled into binaries containing machine code and installs them needs to be
+separately built for every supported platform. Therefore, it is distributed as
+architecture-dependent packages. The Python language interpreter is an example of such software.
+
+A package that installs only data files and interpreted scripts can be made
+architecture-independent. It can be built on any supported platform, and the build should always
+result in the same files being installed, irrespective of platform used to perform it. An example of
+this is a so-called pure Python package, i.e. a distribution that installs `.py` modules and
+no compiled extensions. The equivalent Python packaging concept is a `*-none-any.whl` package.
+
+A special case of this are pure Python packages with entry points. Installed entry points are
+platform-specific: on Unixes they are pure Python scripts, but on Windows they are compiled
+executables. In order to facilitate `noarch: python` packaging for them, entry points are stored not
+as final executables, but as the original list. When the package is installed, they are recreated in
+the appropriate platform-specific format.
+
+Conversely, a Python distribution that installs compiled extension modules in addition to `.py`
+modules needs to be built for every platform separately, and requires using arch-dependent packages.
+Furthermore, since Python extensions interface with the Python interpreter, they also need to be
+concerned about ABI compatibility with it. Python exposes two kinds of ABI: the regular extension
+ABI and the stable ABI.
+
+The regular extension ABI preserves compatibility across patch releases of the Python interpreter,
+but is not compatible across major or minor versions. For example, an extension compiled against
+Python 3.12 cannot be used on 3.11 or 3.13. Appropriately, the respective conda-forge package needs
+to be built not only for all supported platforms, but also as separate variants for all supported
+Python versions. At the time of writing, NumPy is an example of such a package.
+
+The stable ABI, on the other hand, guarantees forward compatibility with future minor releases of
+Python. Therefore, an extension built for the stable ABI of Python 3.12 can be successfully used on
+Python 3.13 and 3.14 (but not 3.11). This is represented by `*-abi3-*.whl` Python packages. In
+conda-forge, such packages are built for the oldest supported Python version, and therefore are
+independent of Python version. However, they are still architecture-dependent. An example of such a
+package is rustworkx.


docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,148 @@
...
+  `lib{name}.1.2.3.dylib`,
+- a partially versioned symbolic link, such as `lib{name}.so.1` or `lib{name}.1.dylib`,
+- an unversioned symbolic link, such as `lib{name}.so` or `lib{name}.dylib`.
+
+When building a new program, the linker uses the unversioned symbolic link to select the library
+to link to. From it, the linker reads the library `SONAME` (or install name, on macOS) that
+corresponds to the partially versioned name, and stores that in the resulting executable. When run,
+the executable loads the library using its partially versioned name. When the library is upgraded
+to a newer version that is backwards compatible (i.e. has the same partially versioned name), all
+executables start using the new version.
```
isuruf: On macOS, there's more checking. specificlly the compatibility version and current version stored in the binaries.

mgorny: Okay, read up on it and tried to explain it. Though I feel like it's gotten a bit complex now.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+3. A linker (or link editor), that is responsible for combining multiple object files together,
+   along with the program's dependent libraries, and outputting the actual binary.
+
+Typically, these programs are not run separately, but rather invoked automatically via the compiler
+frontend. Usually, compiler frontends are invoked by build systems, from simple Makefiles to complex
+solutions such as CMake or Meson. These often split compilation and linking steps, compiling every
+object file separately to enable running multiple compiler processes simultaneously to benefit from
+multiple logical CPUs.
+
+## Build systems
```
h-vetinari: I'd kinda prefer to make a distinction between "Build orchestrators" (CMake etc.) and "Build systems" (ninja etc.). Open to other naming suggestions, as long as we can somehow keep them separate (perhaps with an admission that CMake etc. are often called build systems when that distinction is irrelevant)

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+
+Many programming languages fall somewhere across the compiler-interpreter spectrum. Many interpreted
+languages use Just-in-Time (JIT) compilation: rather than interpreting the input directly, they
+compile it into machine code upon loading, achieving better performance with one-time startup cost.
+Python interpreters commonly compile source code into more efficient bytecode upon loading. On the
+other hand, Java sources need to be manually compiled into bytecode that is afterwards interpreted
+by a Java Virtual Machine.
+
+This document specifically focuses on the concepts related to languages compiled into machine code,
+such as C, C++ and Rust.
```
h-vetinari, line 28: Love the intro!

docs/maintainer/understanding_conda_forge/compilation-concepts.md
h-vetinari, line 48: I'd kinda prefer to make a distinction between "Build system generators" / "Build orchestrators" / "Build management tools" (CMake etc.) and "Build systems" / "Build executors" (ninja etc.).
Open to other naming suggestions, as long as we can somehow keep them separate (perhaps with an admission that CMake etc. are often called build systems when that distinction is irrelevant).

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+   stage.
+3. Testing -- which covers running the package's test suite to verify that is working correctly. In
+   some build system, tests are compiled in this stage; in others, they need to be enabled at
+   configuration stage and built at compilation stage.
+4. Installation -- which covers installing compiled files and package's data files to the actual
+   system, or a staging directory.
+
+Projects usually adapt one of the existing build systems, such as GNU autotools, CMake or Meson.
+These build systems in turn generate Makefiles, Ninja build scripts or project files for IDEs to use
+and perform the actual build.
```
h-vetinari: Let's turn this into a short section
```suggestion
## Build executor

The build orchestrator usually does not execute the actual commands itself, but calls a dedicated
build executor such Ninja, make, etc. Orchestrators may support generating instructions for several
build executors, or different executors per platform. Projects often have a preferred executor.
In almost all cases, build orchestrator and build orchestrator can be considered as if they were
acting as a single unit.
```

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+
+## Libraries
+
+Modern programs almost always link to libraries, that is collections of shared code. This includes
```
h-vetinari:
```suggestion
Modern programs almost always link to libraries, that is, collections of shared code. This includes
```

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+
+Dynamic linking creates programs that reuse a shared copy of the library. As such, the library
+either needs to be installed separately or distributed along the program. However, their main
+advantage is that the same library is shared across multiple packages, and can be quickly swapped
+for another version as necessary. However, this requires one to take special care for different
+library versions to be compatible in the Application Binary Interface (ABI) exposed to programs.
+
+In conda-forge, dynamic linking to libraries provided by conda-forge packages is preferred. Many of
+the concerns related to dynamic linking do not apply here, as packages take care of installing the
+library dependencies in compatible versions.
```
h-vetinari:
```suggestion
In conda-forge, dynamic linking to libraries provided by conda-forge packages is strongly preferred. Many of
the concerns related to dynamic linking do not apply here, as proper packaging practices
ensure that library dependencies are installed in compatible versions.
```

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+
+## Development files
+
+Building against libraries requires additional development files to be available. In conda-forge,
+these files are usually installed as part of the package installing the library itself; in some
+cases they are split to separate "development" packages though.
+
+The necessary development files include:
+
+- header files and include files, providing function prototypes and inline code from the library,
```
h-vetinari, line 170: I was going to note this for the review overall, but "function prototypes" here triggers the response early. Originally when I added the bullet point "compilation concepts", I did want to also explain, symbols, the static/dynamic distinction from the POV of the linker, and ABI breaks.

The reason I think that it'd be good to go to the level of how the dynamic linker works, is that I believe that _this_ is the level at which it becomes evident why run-exports exist and (in conjunction with ABI constraints) how they work. Once you understand that symbols are either copied into the binary from a static lib, or only _referred_ to for a dynamic library, it becomes clear that someone, somewhere has to go find that symbol once you're in the final environment. The run-export ensures that it is present, the linker goes and finds it.

Of course, this field is vast, and not everything needs to be crammed into this PR. However, if you'd like, feel free to crib as much (or as little) as you want from https://pypackaging-native.github.io/background/compilation_concepts, which is where I wrote up a bunch of stuff on this ~3 years ago (note that ABI got a separate page there; we can do the same here if we want).

In any case, what I'd like to avoid is using uncommon terms that are not explained in the text itself. So "function prototypes" kinda falls from the sky here, and would likewise benefit some elucidation.

This could literally be a "C in 2 paragraphs" section that explains how functions are separated into a declaration and an implementation (another good segue towards symbols), and how we need to find the matching headers to a given library in order to be able to correctly use the symbols of that name from the shared library.

In terms of core concepts, mentioning the fundamental file system hierarchy into `/include`, `/lib` and `/bin` also could be explained as a consequence here (which obviously feeds into the below).

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+- static libraries, needed for static linking,
+- shared libraries or import libraries, needed for dynamic linking,
+- pkg-config files or build system-specific files, used to indicate how to build against the
+  library.
+
+Include files are installed into the `include` directory or its subdirectories. They usually use
+`.h` or `.inc` suffix for C. For C++, sometimes names without a suffix are used to follow the
+standard library `#include` scheme.
```
h-vetinari: there's also `.hpp` that's very common for `C++`

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+
+On Unix platforms, executables don't feature any suffix, and common shells only start executables
+when their filename matches the specified command exactly. On Windows, executables commonly use
+`.exe` suffix, and shells account for that. For example, the Python executable will be named
+`python` on Unixes, and `python.exe` on Windows; in both cases, typing `python` will execute it.
+Executables are usually installed into the `bin` directory.
+
+Shared libraries use `lib` prefix on Unixes, and `.so` suffix, except for macOS where they use
+`.dylib` suffix instead. They are installed into the `lib` directory. They often include a version
+string to indicate compatibility between different library versions.
```
h-vetinari: We could mention the term "SOVERSION" here; this could also be profitably referenced against ABI & linkers

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+
+On Windows, shared libraries use `.dll` suffix, and no obligatory prefix. They are installed into
+the `bin` directory, along with programs. There is also no standard filename versioning scheme,
+though many libraries include a version in the filename. The `.dll` files are only used at runtime.
+To build programs against a shared library, an additional import library of `.lib` format must be
+used, which also specifies the `.dll` file to use.
```
h-vetinari: not sure this is the best formulation, happy to iterate
```suggestion
To build programs against a shared library, an additional import library of `.lib` format must be
used, which essentially describes the (visible) content of a `.dll` file to use.
```

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+
+On most systems, loadable modules use the same format as shared libraries. Darwin is an exception
+to that, where loadable modules are "bundles". The recommended suffix for these files is `.bundle`,
+though much software (including Python) uses `.so` instead. They are usually installed into
+tool-specific directories.
```
h-vetinari: If we're talking about osx specifics, do we want to briefly mention frameworks?

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+would be compatible with `>=1.2.0,<2`, but the library would also remain compatible with programs
+compiled against earlier versions. While Linux technically encodes the equivalent of compatibility
+versions in the filename, they aren't strictly enforced.
+
+## Finding shared libraries at runtime
+
+Typically, a binary linked to a shared library does not embed the complete path to the library, but
+rather the library name. When a program is started, the dynamic loader is responsible for finding
+all the needed libraries, recursively, and loading them. The exact behavior differs from platform to
+platform. Appropriately, the ways conda-forge builds binaries account for these differences.
```
h-vetinari, line 299: Ah, I should have read further; this section is great!

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+Conda splits package dependencies into three classes:
+
+- `build` -- indicating dependencies that need to be available while building the package, and that
+  are compatible with the build platform. This primarily involves programs being run at build time,
+  but also "sysroot" and CDT packages.
+- `host` -- indicating dependencies that need to be available while building the package, and that
+  are compatible with the host platform. This primarily involves dependent libraries, but also
+  interpreters like `python`.
+- `run` -- indicating dependencies that need to be installed in the system along with the package.
+  This primarily involves programs and libraries used at runtime.
```
h-vetinari: TBH, I think this topic is worth a separate article (this is what I had meant with "Environment roles" in the initial draft of the doc separation). I'd like to remove it here and move that to a separate PR. We can then crossreference the two in whichever PR comes second

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+There are also cases when run exports are undesirable. For example, if a library is linked
+statically or only used to build tests. In these cases, there is also a mechanism to ignore run
+exports, and prevent the dependency from being added to `run` dependencies.
+
+When packages are being built, both conda-build and rattler-build perform linkage analysis on the
+built binaries. They list all the libraries that every binary links against, and map them to the
+respective conda-forge packages. This output can cross-referenced to the final package `run`
+dependencies, to check for missing dependencies (usually belonging in `host` class).
+
+## Native builds, cross builds and emulation
```
h-vetinari: Even though this clearly fits with "compilation concepts", I'd like to move the bulk of this to the environment discussion (and then refer to that from here).

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+languages such as Python or shell script tend to be run via an interpreter that processes and runs
+them directly from the source code. Both approaches have their advantages: compiling into machine
+code can produce efficient stand-alone executables, at the cost of portability; on the other hand,
+an intepreter can run programs straight from sources and handle code changes immediately, at the
+cost of worse performance.
+
+Strictly speaking, programming languages aren't bound into compiler-interpreter dichotomy. There do
+exist interpreters for programming languages like C, and there do exist compilers that can turn
+Python code into stand-alone native executables. Nevertheless, they primarily follow a single model
+and packagers rarely need to be concerned about the alternative possibilities.
```
h-vetinari, line 18: We could revisit the interpreter aspect towards the end of the article; in order to contrast again:
* the python interpreter is compiled per platform
* once you have that, pure python code does not need to be built per platform anymore (--> noarch)
* python packages that want to use native code (like numpy) still need to be built per platform
* where those python packages want to talk to the python interpreter directly, they are exposed to python's own ABI, and thus need to be build per python version (unless when using the limited API/stable ABI --> `python_version_independent:)

mgorny:
> For the environment stuff (+cross-compilation) I have a certain pedagogy in mind, if you want I can prepare a draft of that based on what you wrote up here. I'd like to split that out though

Note that there's also https://conda-forge.org/docs/maintainer/understanding_conda_forge/cross-compilation/ with some overlap.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
mgorny: To be honest, I wasn't sure about the best ordering. I've kinda tried to keep the order so that you could read top-to-bottom without getting too many "to be explained later" things.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+and perform the actual build.
+
+## Libraries
+
+Modern programs almost always link to libraries, that is collections of shared code. This includes
+system libraries and compiler runtimes that provide the basic programming language functionality,
+and third-party libraries provided by various packages.
+
+Programs link to libraries either statically or dynamically. Static linking means that the library
+code is embedded into the program directly, and the library is no longer needed at runtime. Dynamic
```
jaimergp:
```suggestion
code is embedded into the program directly, and the library _file_ is no longer needed at runtime. Dynamic
```
A small clarification in case someone interprets this as the "library (logic, code) is not used at runtime".

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+linking means that the program merely carries a reference to an external library, and the library is
```
jaimergp:
```suggestion
linking means that the program merely carries a reference to an external library, and the library file is
```

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+loaded when the program starts. Both approaches to linking have their use cases, and their
+proponents.
+
+Static linking creates a standalone program that is easier to distribute, and may benefit from
+additional optimization as the optimizer is able to determine how the library is used exactly.
+However, statically linked programs are less space efficient, especially if the same library is used
+across multiple programs. Since a specific library version is embedded into the program, the risk of
+breakage on updates is minimized. However, this means that in order to update the library, the whole
+package needs to be rebuilt, which may negative impact security response time if the library turns
```
jaimergp:
```suggestion
package needs to be rebuilt, which may negatively impact security response time if the library turns
```

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+out to be vulnerable.
+
+Dynamic linking creates programs that reuse a shared copy of the library. As such, the library
+either needs to be installed separately or distributed along the program. However, their main
```
jaimergp:
```suggestion
either needs to be installed separately or distributed along with the program. However, their main
```

Not sure about this one.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+advantage is that the same library is shared across multiple packages, and can be quickly swapped
+for another version as necessary. However, this requires one to take special care for different
```
jaimergp:
```suggestion
for another version as necessary. Unfortunately, this requires one to take special care for different
```

To avoid repetition of "However".

docs/maintainer/understanding_conda_forge/compilation-concepts.md
jaimergp: I would even add that "proper packaging practices ensure that library dependencies are _annotated_ and installed in compatible versions".

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+
+In conda-forge, dynamic linking to libraries provided by conda-forge packages is preferred. Many of
+the concerns related to dynamic linking do not apply here, as packages take care of installing the
+library dependencies in compatible versions.
+
+## Development files
+
+Building against libraries requires additional development files to be available. In conda-forge,
+these files are usually installed as part of the package installing the library itself; in some
+cases they are split to separate "development" packages though.
```
jaimergp: Can we say here that conda-forge packages _may_ or _may not_ ship the development files in the same artifact, depending on criteria such as their complexity, popularity and size? Now it sounds like we almost always ship them together intentionally, when most of the time is out of convenience, simplicity and... laziness :)

docs/maintainer/understanding_conda_forge/compilation-concepts.md
jaimergp: Some hyperlinks and references may help here too, in case an advanced concept needs to be explained separately (e.g. by linking to some official documentation somewhere).

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+Include files are installed into the `include` directory or its subdirectories. They usually use
+`.h` or `.inc` suffix for C. For C++, sometimes names without a suffix are used to follow the
+standard library `#include` scheme.
+
+Static libraries are installed into the `lib` directory and carry an `.a` suffix on Unix, or `.lib`
+on Windows. On Unix, shared libraries are used directly for linking, and they are described in the
+[binaries](#binaries) section. On Windows, import libraries are used instead. They use `.lib` suffix
+like static libraries.
+
+Finally, packages often provide additional files that are used at build time to determine how to use
```
jaimergp:
```suggestion
Finally, packages often provide additional files that are used at build time to determine how to compile against
```

Is this too specific? "Use" here could be understood as "execute library code".

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+2. Shared libraries: collections of compiled code that programs usually link to and therefore they
+   are loaded by programs at start time.
+3. Loadable modules: collections of compiled code that are loaded by programs at runtime (e.g.
+   plugins).
+
+On Unix platforms, executables don't feature any suffix, and common shells only start executables
+when their filename matches the specified command exactly. On Windows, executables commonly use
+`.exe` suffix, and shells account for that. For example, the Python executable will be named
+`python` on Unixes, and `python.exe` on Windows; in both cases, typing `python` will execute it.
+Executables are usually installed into the `bin` directory.
```
jaimergp:
```suggestion
Executables are usually installed into the `bin` directory.
```

In Unix. Windows is... weirder. See [CEP PR#133](https://github.com/conda/ceps/pull/133) for some examples.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+
+Shared libraries use `lib` prefix on Unixes, and `.so` suffix, except for macOS where they use
```
jaimergp:
```suggestion
Shared libraries use filenames with a `lib` prefix on Unixes, and `.so` suffix, except for macOS where they use
```

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+`.dylib` suffix instead. They are installed into the `lib` directory. They often include a version
+string to indicate compatibility between different library versions.
+
+On Windows, shared libraries use `.dll` suffix, and no obligatory prefix. They are installed into
+the `bin` directory, along with programs. There is also no standard filename versioning scheme,
```
jaimergp: Again, not always. I would say "along with the executable programs (usually under `bin/` or equivalent)".

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+though many libraries include a version in the filename. The `.dll` files are only used at runtime.
+To build programs against a shared library, an additional import library of `.lib` format must be
+used, which also specifies the `.dll` file to use.
+
+On most systems, loadable modules use the same format as shared libraries. Darwin is an exception
```
jaimergp:
```suggestion
On most systems, loadable modules use the same format as shared libraries. macOS is an exception
```

For consistency with the terminology used so far.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+to that, where loadable modules are "bundles". The recommended suffix for these files is `.bundle`,
+though much software (including Python) uses `.so` instead. They are usually installed into
+tool-specific directories.
+
+## Shared library versioning
+
+Shared libraries are often versioned to indicate compatibility. Typically, at least two version
+components are used: a minor version that is incremented whenever backwards-compatible ABI changes
+occur (e.g. new interaces are added), and a major version that is incremented whenever
```
jaimergp:
```suggestion
occur (e.g. new interfaces are added), and a major version that is incremented whenever
```

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+backwards-incompatible changes happen. Often additional version components are used to indicate
+library updates without ABI changes.
+
+When such a scheme is used, the installed library usually consists of three files:
+
+- the actual library with a full version string, such as `lib{name}.so.1.2.3` or
+  `lib{name}.1.2.3.dylib`,
+- a symbolic link including the major version, such as `lib{name}.so.1` or `lib{name}.1.dylib`,
```
jaimergp:
```suggestion
- a symbolic link including only the major version, such as `lib{name}.so.1` or `lib{name}.1.dylib`,
```

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+- an unversioned symbolic link, such as `lib{name}.so` or `lib{name}.dylib`.
+
+When building a new program, the linker -- if passed `-l{name}` -- uses the unversioned library
+name. If it is a symbolic link, it is resolved to the actual library. That library is used during
+the linking process, and dynamic library information is taken from it.
+
+On Linux, this information primarily involves the `DT_SONAME` entry that contains the filename used
```
jaimergp:
```suggestion
On Linux, this information primarily involves the `DT_SONAME` metadata entry that contains the filename used
```

What type of entry? Library metadata?

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -0,0 +1,318 @@
...
+directories are searched for dependent libraries (provided they do not specify a full path):
+
+1. The directories specified in the `DT_RPATH` entry of the program, provided `DT_RUNPATH` is not
+   present. Specifying `DT_RPATH` is discouraged, since the resulting behavior is suboptimal.
+2. The directories specified in the `LD_LIBRARY_PATH` environment variable. This variable is
+   typically set locally when library search paths need to be overridden.
+3. The directories specified in the `DT_RUNPATH` entry of the binary. Note that these paths do not
+   apply recursively -- the program's `DT_RUNPATH` is used for the libraries used directly by the
+   program, and these libraries's entries are used for their own dependent libraries, and so on.
+4. The standard system search paths.
```
jaimergp, line 312: There are specified somewhere, right? Should we talk about `ld.so.conf` and friends or is that to advanced?

docs/maintainer/understanding_conda_forge/compilation-concepts.md
jaimergp: +1

jaimergp:
Such a great read, @mgorny! Thanks for the write up. I left some comments as a non-expert who has only had limited contact with this contact through (intuition-driven at best) practice. 🙏

docs/maintainer/understanding_conda_forge/compilation-concepts.md
mgorny: I've assumed we want to skip the gory details that aren't that relevant to conda-forge, but I can expand on that. The main point was that these things get quite complex on Windows, and I really didn't want to copy the huge lists from MSDN here, so kept others at roughly the same level of detail.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
mgorny: Yeah, it's a complex topic. That said, I'd avoid uncommon terms like "orchestrators"; if only because this would mean we're inventing new nomenclature (never heard anyone talking about "build orchestrator".

Overall, I've tried to bunch them all together because there are many different ways in which people use them: a Makefile can either be a simple output of a "generator", or a full-fledged build system with lots of GNU extensions to make it work. Also, at least `cmake` and `meson` seem to be aiming towards hiding the underlying builder as an implementation detail: `cmake --build`, `meson compile`).

docs/maintainer/understanding_conda_forge/compilation-concepts.md
mgorny: I'm not sure if "metadata" is the right term. It's part of the dynamic section, so I suppose it's more like data, though I'm not sure if we really want to go deep into the gory details here.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
mgorny: I see that CMake and Meson both call themselves a "build system". GNU makes [describes itself](https://www.gnu.org/software/make/) as a "tool which controls the generation of executables and other non-source files of a program from the program's source files." Wikipedia seems to arbitrarily talk of "software development tools", "build automation tools".

docs/maintainer/understanding_conda_forge/compilation-concepts.md
mgorny: Oh, that's nice. I suppose we could actually integrate most of it. Luckily, it doesn't seem to overlap much with what I wrote.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
mgorny: Done.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
mgorny: Well, I suppose "usually" holds then? :-)

I don't think that's the right place to try to explain all of this. Maybe I shouldn't have started listing the directories in the first place.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
```diff
@@ -93,6 +93,45 @@ Note that in some cases the header files may contain implementation in addition
...
+strict contract with the program that an `int` value must be passed.
+
+Now consider that the library changes prototype into:
+
+```c
+void foo(int64_t a);
+```
+
+From programmer's perspective this can be fine, as long as the previous `int` input can be converted
+into `int64_t`. However, the binary contract changes -- a previously compiled program passes an
```
h-vetinari: Let's not rely on wobbly-sized `int`egers in the explanation. Aside from the fact that those _can_ be 64bits under ILP, we shouldn't have to rely on implicit knowledge in such an explanation. From my perspective it'd be good enough to just use `int32_t` for the first example.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
mgorny: Hmm, the file system hierarchy part is definitely interesting to include, though I wonder if this is actually the best article to put it into. Perhaps it would make sense to have it somewhere else, and link from here.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
mgorny: Hmm, that probably makes sense. I need to read up on them.

docs/maintainer/understanding_conda_forge/compilation-concepts.md
mgorny: This is definitely a nice idea. I need to think how to word it, though.

jaimergp (approved):

PR #2661, opened 2025-11-21
How to: Maintain several versions

jaimergp:
PR Checklist:

- [ ] note any issues closed by this PR with [closing keywords](https://help.github.com/articles/closing-issues-using-keywords)
- [X] if you are adding a new page under `docs/` or `community/`, you have added it to the sidebar in the corresponding `_sidebar.json` file
- [ ] put any other relevant information below

Closes https://github.com/Quansight-Labs/conda-ecosystem-sta-mgmt/issues/63

PR #2711, opened 2026-01-07
add instructions for how to branch off at commits before current HEAD

h-vetinari:
Follow-up to #2661, for an important case that I had missed back then. CC @mgorny @jaimergp

### DIFF
--- a/docs/how-to/advanced/several-versions.md
+++ b/docs/how-to/advanced/several-versions.md
@@ -26,8 +26,16 @@ The above is for branching off from the `main` branch directly; in many cases th
 to branch off at an earlier point of the repo history. For example, if the `main` branch is already
 on version 3.11 or beyond, but you want to publish a new patch release 3.10.18. In this case, you
 should go `https://github.com/conda-forge/<your_pkg>-feedstock/commits/main/` and find the commit
-for the last _published_ builds of the 3.10.x series (usually: the last merge before any of the
-commits from the PR that did the update to 3.11). Then, instead of `git checkout -b v3.10.x`, do:
+for the last _published_ builds of the 3.10.x series -- in other words, a commit where there is some
+indication that the CI ran successfully (green check or, more rarely, a red X where _some_ jobs still
+succeeded). Usually this is the last merge before any of the commits from the PR that did the update
+to 3.11.
+
+Alternatively, you can go to the conda-forge [metadata app](https://conda-metadata-app.streamlit.app),
+navigate to the last 3.10.x build of your package, and look for the "Provenance" section. It contains
+a reference to the exact commit that produced the package you're looking at.
+
+Once you've found the right commit hash, instead of the `git checkout -b v3.10.x` described above, do:
 
 ```bash
 git checkout <sha_hash_of_commit_from_above> -b v3.10.x
@@ -55,10 +63,10 @@ file and add these lines:
 ```yaml
 bot:
   abi_migration_branches:
-    - "v3.10.x"  # the branch name you picked
+    - "v3.10.x"  # the branch name you picked; use a consistent scheme across versions
 ```
 
-You can incorporate this in your next PR to the `main` branch, or do a push with
+You can incorporate this change into your next PR to the `main` branch, or do a push with
 
 ```bash
 git checkout main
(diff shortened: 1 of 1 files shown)


jaimergp (approved):

docs/how-to/advanced/several-versions.md
```diff
@@ -22,34 +22,60 @@ git pull upstream main
 git checkout -b v3.10.x
 ```
 
-Add an empty commit with the `[ci skip]` message so the new branch does not result in a new build process, and push it to `upstream`. This is one of the rare ocassions where you must push directly to the feedstock, not your fork!
+The above is for branching off from the `main` branch directly; in many cases though you may want
+to branch off at an earlier point of the repo history. For example, if the `main` branch is already
+on version 3.11 or beyond, but you want to publish a new patch release 3.10.18. In this case, you
+should go `https://github.com/conda-forge/<your_pkg>-feedstock/commits/main/` and find the commit
+for the last _published_ builds of the 3.10.x series (usually: the last merge before any of the
```
mgorny: I wonder if we should hint that one should look at CI status marks.

docs/how-to/advanced/several-versions.md
```diff
@@ -22,34 +22,60 @@ git pull upstream main
...
 ```
 
-Now, this branch can be selected as a target branch in the following steps.
-
-## Open a PR with the necessary changes
-
-From the same branch, create another one to add some extra changes that will need to be reviewed in a PR:
+In either case, add an empty commit with the `[ci skip]` message so that the act of pushing the new
+branch does not result in a new build process, and push it to `upstream`. This is one of the rare
+ocassions where you must push directly to the feedstock, not your fork!
```
mgorny:
```suggestion
ocassions where you must push directly to the feedstock repository, not your fork!
```

docs/how-to/advanced/several-versions.md
```diff
@@ -22,34 +22,60 @@ git pull upstream main
...
 ```
 
-And [rerender](../basics/rerender.md). Now, make sure to adjust the recipe file so the correct version is being built.
+Note that the bot infrastructure for `abi_migration_branches:` only does migrations; it will not
+automatically pick up new patch versions, which you have to do manually (e.g. by subscribing to
+notifications for new releases in the upstream repo).
 
-Once ready, push the branch to your fork (`origin`) and open the corresponding pull request. Don't forget to pick `v3.10.x` as the target branch!
+Eventually, you'll want to stop receiving bot PRs to a given branch (because the version has become
+too old); at that point, you simply remove the branch from `abi_migration_branches:` again. Again, the
```
mgorny:
```suggestion
too old); at that point, you simply remove the branch from `abi_migration_branches:`. Again, the
```

Double "again", also it sounds weird, given you didn't really remove it before.

docs/how-to/advanced/several-versions.md
```diff
@@ -22,34 +22,60 @@ git pull upstream main
...
-    - "v3.10.x"  # or the branch name you picked
+    - "v3.10.x"  # the branch name you picked
+```
+
+You can incorporate this in your next PR to the `main` branch, or do a push with
+
+```bash
+git checkout main
+git pull upstream main
+git add conda-forge.yml
```
mgorny, line 75: I suppose it's fine in this context, but still feels a bit wrong to be editing the file before changing branches.

docs/how-to/advanced/several-versions.md
h-vetinari: we can. I assumed that "last merge" was also the last CI run, which is true in ~99% of the cases. Do you have a suggestion for a better formulation?

docs/how-to/advanced/several-versions.md
h-vetinari: OK, I see what you mean; I had meant that the edit comes between `git pull` and `git add`, but didn't note that in the workflow...

h-vetinari:
```suggestion
git checkout main
git pull upstream main
# modify conda-forge.yml and save
git add conda-forge.yml
```

docs/how-to/advanced/several-versions.md
mgorny: Hmm, maybe "the last commit annotated with CI tick mark …"?

tjwilli58: As a user that will be needing this soon, can you explain how to find these CI tick marks?

jaimergp: In the `/main/commits` page of your feedstock; e.g. https://github.com/conda-forge/conda-rattler-solver-feedstock/commits/main/

image

jaimergp: Another way of checking which commit was responsible for building a given version is to check the provenance details in conda-metadata-app, e.g. in https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Fnumpy-2.4.1-py314hd4f4903_0.conda

image

h-vetinari: Thanks for the inputs here! I've fleshed out the text some more - PTAL! :)

PR #2723, opened 2026-01-22
Make name the second key by default in sorting migration table

### LINKED ISSUE #2720: Fix sorting of migration tables
Hey @Carreau; your work on this has been awesome and is much appreciated! I'm not deep in typescript and not the right person to review here, but I wanted to note a minor regression in all the work so far, which is that feedstocks were previously ordered by `(num_children,name)`; the sorting by num_children still works, but alphabetization has gone

image

_Originally posted by @h-vetinari in https://github.com/conda-forge/conda-forge.github.io/issues/2670#issuecomment-3710026802_

Carreau:
Closes #2720

This restore the previous default.

### DIFF
--- a/src/pages/status/migration/index.jsx
+++ b/src/pages/status/migration/index.jsx
@@ -483,7 +483,7 @@ function Table({ details }) {
   const CI_STATUS_ORDER = { clean: 0, behind: 1, has_hooks: 2, unknown: 3, unstable: 4, blocked: 5, dirty: 6, draft: 7, "": 999 };
 
   // Transform data to match expected structure for compare function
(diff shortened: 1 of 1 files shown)


h-vetinari (approved):
Thanks a lot for the quick fix! 🙏

h-vetinari:
> No worries; I can understand why i can be annoying, I have no idea how to have a better sorting interface though.

There's definitely a "t" missing there. You're not annoying, you're awesome! :)

I agree it'd be nice to have a better sorting interface, but as long as the default is sane, that's secondary IMO.

Issue #2164, opened 2024-04-30
RFC: restructure website content, roughly following Diátaxis


h-vetinari:
There are several ongoing efforts to improve the conda-forge website, which has yielded the new website, status page, as well as some work around the infrastructure pages.

Additionally @jaimergp [wrote](https://github.com/conda-forge/conda-forge.github.io/issues/2154#issuecomment-2064124108) in the context of the more tightly-scoped #2154:
> Last year, I experimented with how Diátaxis would apply to our docs in https://cf-infra-docs.netlify.app/docs. Some of the docs written for that experiment are being migrated over to our new website. @zklaus is almost finished, I think, with that effort.

I took that idea and came up with a rough draft of how the website could be structured (originally looking only at the infrastructure page, but I felt I needed a big picture view even for that). Some new content would have to be written, but mainly it is intended to reshuffle existing content into easily discoverable and more granular pieces.

As [Diátaxis](https://diataxis.fr/) intentionally aims at gradual transformation rather than big bang refactoring, this outline would likely change as we iterate, but I wanted to get people's thoughts on the rough strokes (and then discuss it in the core call), so we can agree on the overall direction and let people start chipping away at some first pieces.

This is based on looking at our current docs as well as @jaimergp's [experiment](https://cf-infra-docs.netlify.app/docs) with Diátaxis; a more detailed comparison is available below the fold.

```
Welcome
For users
└ ...                               # not in scope here
For maintainers
├ Tutorials
|  ├ Generate a conda recipe        # Using greyskull or whatever
|  ├ Write your first recipe        # Segue from first tutorial, make some changes
|  ├ Submit your first recipe       # Segue again, process for submission
|  └ ...
├ How-to guides
|  ├ Easy                           # the below is just a selection, obviously
|  |  ├ Read a recipe               # Component overview, refer to in-depth explanation
|  |  ├ Rerender feedstock
|  |  ├ Test packages
|  |  ├ Deal with numpy
|  |  ├ Enable osx-arm64            # Same for arches
|  |  ├ Talk to the bots            # refer to bot command reference page
|  |  └ ...
|  ├ Medium
|  |  ├ Cross-compile
|  |  ├ Enable CUDA
|  |  ├ Multi-output recipes
|  |  ├ LTS branches
|  |  └ ...
|  └ Emergencies
|     ├ Patch repodata
|     └ Mark a package broken
├ Understanding conda-forge         # Explanation part of Diátaxis
[...]

Issue #1972, opened 2023-07-02
Branch deletion policy?


h-vetinari:
I was wondering if we have a policy for deleting branches on a repo, and if not, if we should formulate one, at least in rough strokes.

It's clear that these branches are producing builds, so for auditability and traceability, we generally shouldn't delete things, at least not very eagerly.

But some feedstocks where we keep maintenance branches for older versions have amassed a lot of branches, many of which haven't been built in years, and in such cases, I cannot see what harm it could do to delete them. Additionally, occasionally there are branches that get pushed to a repo +/- accidentally, but remain lying around as cruft.

There's a technique that some repos use when deleting a branch, to leave a tag for the head of that commit. That way, it remains easily discoverable for those who need it.

Thoughts?

h-vetinari:
> Are there situations where they become burdensome?

Yes, very much so, especially on feedstocks with many maintenance branches. This is exacerbated by github ordering the branches in alphabetical (rather than reverse chronological) order:

<img src="https://github.com/conda-forge/conda-forge.github.io/assets/33685575/4c2024bd-214c-4766-9179-89e02c08d6f6" width=40% height=40%>

All of these branches are useless nowadays and changing between the _current_ branches for maintenance purposes becomes more of a hassle.

> I'd be inclined to leave them as-are for posterity.

Hence the option to turn them into tags.

I think cleaning up unused branches is part of regular repo hygiene, even for maintenance branches, but doubly so for random branches.

PR #2730, opened 2026-01-29
Document backport branch creation for older releases

jacobtomlinson:
While recently making a backport release for a security fix I failed to find the [How to maintain several versions](https://conda-forge.org/docs/how-to/advanced/several-versions/) page when searching for instructions.

I wrote up a [blog post on making backports on Conda Forge Feedstocks](https://jacobtomlinson.dev/posts/2026/backporting-fixes-to-older-releases-on-conda-forge/) and someone pointed the page out to me.

This PR adds some working which will hopefully make it more searchable for other folks making backports in the future.

PR Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,4 +1,4 @@
-name: deploy
+name: Website
 
 on:
   push:
@@ -8,22 +8,31 @@ on:
   workflow_dispatch: null
 
 jobs:
-  tests:
-    name: deploy
-    runs-on: "ubuntu-latest"
+  build:
+    name: Build
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
-
-      - uses: prefix-dev/setup-pixi@82d477f15f3a381dbcc8adc1206ce643fe110fb7 # v0.9.3
-
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: prefix-dev/setup-pixi@a0af7a228712d6121d37aba47adf55c1332c9c2e # v0.9.4
       - name: build
         run: pixi run build-production
         env:
           GHREF: ${{ github.ref }}
           GHREPO: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
+      - uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+        with:
+          path: build/
 
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-slim  # 1xCPU
+    needs: [build]
+    if: github.ref == 'refs/heads/main' && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
+    steps:
+      - uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
+        with:
+          path: build/
       - name: deploy
-        if: github.ref == 'refs/heads/main' && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
         uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4.0.0
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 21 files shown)


docs/how-to/advanced/several-versions.md
```diff
@@ -57,7 +59,11 @@ a new patch version of v3.10.x in the example above).
 
 For the duration of the lifetime of your LTS version, you'll want it to participate in the relevant
 migrations that are happening in conda-forge, e.g. if there's a rebuild for a new version of one of
-your dependencies. The configuration for this must be on the `main` branch in the `conda-forge.yml`
+your dependencies.
+
+If you're making a one-off backport release you can skip this section.
```
jakirkham: So this may or may not be true depending on the use case

In our case Dask doesn't really get ABI migrations as a pure Python package. So it doesn't really matter

Usually this comes up when a package has compiled code and links to other library dependencies. For example `libcurl`. Then if `libcurl` is updated to a new version that breaks ABI, the conda-forge stack is migrated to the new version. The bot will make PRs to update the `libcurl` to all the branches in this configuration list. If the list is empty, only `main` will see updates

If the hot-fix release needed to be maintained for some reason, then maybe this could make sense

docs/how-to/advanced/several-versions.md
```diff
@@ -8,6 +8,8 @@ The conda-forge workflow assumes that a push to any branch in the feedstock repo
 
 Most feedstocks only need `main` for their builds, since the package has a single release line, and new releases always imply a later version. However, some packages may maintain a few release lines in parallel. If you wish to maintain those in your feedstock, you will need to create a branch for each.
 
+There also might be occasions where you want to backport a bug fix or security patch to an older release. In these situations you may want to make a branch specifically for the backport version you want to release.
```
h-vetinari: As a blanket statement, I'm not really comfortable with this. The branching strategy should be as _coarse_ as feasible[^1], though exceptions can of course occur.

Going by the 3.10.x vs. 3.11 example, let's assume that 3.10.18 is the last release in the 3.10 line, and that the security issue exists since 3.10.14. In this case, you should _not_ create a 3.10.14, 3.10.15, etc. branch. Rather, users should move to the latest patch version (that's one of the main roles of patch versions! no features, thus basically no breakage, but definitely receiving relevant security updates). In other words, the `3.10.x` branch would suffice here.

[^1]: Partially, that's because branches must not be deleted (which is actually something we need to document on this page), because it destroys traceability of artefacts who were published from that branch. However we don't have a way to clean them up yet, so creating lots of branches creates a lot of clutter. Eventually I hope we'll get around to implement the delete-branch-by-turning-it-into-a-tag described in #1972, then we could also protect all branches by default.

docs/how-to/advanced/several-versions.md
h-vetinari: Like John says, this depends on whether your package has ABI-dependencies that get migrated. If so, the rest of conda-forge will move on, while your package's dependencies effectively stay frozen in time due to lack of compatible builds with newer baseline versions.

So even for a one-off backport, the long-term usability of that depends on enabling migrations. It can be skipped when no dependency is affected by our migrations, but that's pretty hard (though not completely impossible) to determine by hand.

h-vetinari (changes requested):
Thanks for the PR. We haven't really addressed the discoverability issue that caused you to miss the page in the first place. What would have helped you find this page better? 

I fully admit that it's not ideal. The efforts towards #2164 are happening a bit behind the scenes (#2633) as we slowly fill it with content, so that definitely doesn't help discoverability. But we're also not at a point yet where we can switch over I think. Perhaps we can still find a way to make it more visible in the meantime. CC @jaimergp

docs/how-to/advanced/several-versions.md
```diff
@@ -8,6 +8,8 @@ The conda-forge workflow assumes that a push to any branch in the feedstock repo
 
 Most feedstocks only need `main` for their builds, since the package has a single release line, and new releases always imply a later version. However, some packages may maintain a few release lines in parallel. If you wish to maintain those in your feedstock, you will need to create a branch for each.
 
+There also might be occasions where you want to backport a bug fix or security patch to an older release. In these situations you may want to make a new branch off from the old release you want to backport to so that you can continue making new releases there.
```
h-vetinari: The "there might be occasions" is giving a false impression that this is a side goal. It is the _main_ goal of this document. So I propose to reframe this. Something along the lines of.

```suggestion
Important reasons why it can be necessary to keep publishing packages for older versions:
* The package continues to receive upstream support, and new patch versions need to be published.
* The package depends on libraries that are versioned in the global pinning, which need to be rebuilt by our bot infrastructure to stay compatible with the rest of conda-forge.
* The package needs a backport of an upstream patch, security fix, etc.
```

h-vetinari (changes requested):
OK, if mentioning the word "backport" is enough for visibility (🤞), I'm fine with that. Here's a suggestion how to do that.

h-vetinari (approved):
Thanks!

PR #2864, opened 2026-06-18
FIX: Add closing quote to match condition

matthewfeickert:
* Add missing `"` to close the match condition.
* As setuptools is not the only build system that exists after PEP 517/518 use the term "build system" and then provide setuptools as an example.

PR Checklist:

- [N/A] note any issues closed by this PR with [closing keywords](https://help.github.com/articles/closing-issues-using-keywords)
- [N/A] if you are adding a new page under `docs/` or `community/`, you have added it to the sidebar in the corresponding `_sidebar.json` file

This PR addresses multiple things on a single docs page. If it is preferable to split this between multiple PRs I can split out the doc improvements (`--no-deps --no-build-isolation` and build systems) from the syntax fix.

### DIFF
--- a/docs/maintainer/adding_pkgs.mdx
+++ b/docs/maintainer/adding_pkgs.mdx
@@ -393,7 +393,7 @@ You can e.g. specify not to build …
 
   ```recipe
   build:
-      skip: match(python, "<3.12)
+      skip: match(python, "<3.12")
   ```
   </RecipeTabs>
 
@@ -454,8 +454,8 @@ These options should be used to ensure a clean installation of the package witho
 dependencies. This helps make sure that we're only including this package,
 and not accidentally bringing any dependencies along into the conda package.
 
-Usually pure-Python packages only require `python`, `setuptools` and `pip`
-as `host` requirements; the real package dependencies are only
+Usually pure-Python packages only require `python`, a build system (e.g. `setuptools`),
+and `pip` as `host` requirements; the real package dependencies are only
 `run` requirements.
 
 <a id="requirements"></a>


matthewfeickert (commented):
Relevant build view: https://deploy-preview-2864--conda-forge-previews.netlify.app/docs/maintainer/adding_pkgs/#skipping-builds

docs/maintainer/adding_pkgs.mdx
```diff
@@ -415,7 +415,7 @@ The default names for the build scripts are
 
 If the build can be executed with one line, you may put this line in the
 `script` entry of the `build` section of the recipe file with:
-`script: "{{ PYTHON }} -m pip install . -vv"`.
+`script: "{{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation"`.
```
xhochy: I thought this was set automatically? So there is no need to do it manually.

matthewfeickert: @xhochy I believe this is correct for `rattler-build`, but I do not know if this is true for `conda-build`. I had suggested removing these flags before in https://github.com/conda/grayskull/issues/582, but @ocefpaf has requested that they remain in (c.f. https://github.com/conda/grayskull/issues/582#issuecomment-2482370940).

Regardless, the documentation should be consistent across the conda-forge docs. So once choice should be selected and then used everywhere.

ocefpaf: This fails only on multiple outputs and conda-build as far as I know. That is where my recommendation comes from. If we ditch v0, and rattler-build multiple outputs does suffer from that. I agree, don't include it b/c it is unnecessary. Regarding consistency... I'll abstain from that bike shed ;-p

matthewfeickert: > This fails only on multiple outputs and conda-build as far as I know.
> ...
> I agree, don't include it b/c it is unnecessary.

If there is agreement that for v1 recipes this is redundant and should be avoided, should we keep the `--no-deps --no-build-isolation` flags in below for the `v0` example?

docs/maintainer/adding_pkgs.mdx
```diff
@@ -431,7 +431,7 @@ Normally Python packages should use this line:
 
 ```recipe
 build:
-  script: "{{ PYTHON }} -m pip install . -vv"
+  script: "{{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation"
```
matthewfeickert: As this is the v0 recipe instructions, keep the flags given https://github.com/conda-forge/conda-forge.github.io/pull/2864#discussion_r3437751658

docs/maintainer/adding_pkgs.mdx
matthewfeickert: I moved further discussion to Issue https://github.com/conda-forge/conda-forge.github.io/issues/2869

matthewfeickert:
Reminder ping to @conda-forge/core for review.

docs/maintainer/adding_pkgs.mdx
jaimergp: This was addressed in conda-build 24.7.0 I believe? https://github.com/conda/conda-build/blob/main/CHANGELOG.md#2470-2024-07-18

@isuruf fixed it via https://github.com/conda/conda-build/pull/5368

docs/maintainer/adding_pkgs.mdx
jaimergp: See my reply in that thread, it was addressed in 24.7.0

jaimergp:
```suggestion
  script: "{{ PYTHON }} -m pip install . -vv"
```

hmaarrfk (approved):

PR #2882, opened 2026-07-09
Split and update the "noarch" howto

mgorny:
PR Checklist:

Split the "noarch" part of KB into a separate "How to create noarch packages", and update its contents.


mgorny:
This one's ready. I've done some rewriting to update and improve it, and also based on my own misunderstandings.

The two significant changes are:
- I've replaced the wall of notes with an enumerated list and a section dedicated to `noarch_platforms`
- I've removed the point about downstream testing, since according to my local testing it isn't true / up-to-date

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,340 @@
...
+</RecipeTabs>
+
+```yaml title="conda-forge.yml"
+noarch_platforms:
+  - linux_64
+  - osx_64
+  - win_64
+```
+
+Again, remember to rerender after adding / modifying these files so the changes are applied.
```
jaimergp: I'd argue this is not a "basic" how-to and perhaps should be moved to Advanced.

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,340 @@
+---
+tags: [howto, basic]
```
jaimergp:
```suggestion
tags: [howto, basic, noarch]
```

In case we want to also add `noarch` to the proposed advanced split.

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,340 @@
...
+A non-`noarch` package needs to be built for all supported platforms, and for every platform a separate variant is built for every supported Python version.
+This can lead to building over 20 packages that would be roughly equivalent for a pure Python package.
+With `noarch: python`, the same recipe can produce a single package that works across all platforms and Python versions.
+
+In order to qualify as a `noarch: python` package, all of the following criteria must be fulfilled:
+
+- No compiled extensions
+- No post-link or pre-link or pre-unlink scripts
+- No OS-specific build scripts
+- No requirements specific to Python version
```
jaimergp: Let's add an internal note (html comment) about the upcoming repodata v3 conditional dependencies, which will need to be discussed here.

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,340 @@
...
+- `scripts` argument in `setup.py` is not used
+- If `console_scripts` entry points are used (`[project.scripts]` in `pyproject.toml`, `entry_points` → `console_scripts` in `setup.py` or `setup.cfg`), they are also [listed](https://conda.io/projects/conda-build/en/stable/resources/define-metadata.html#python-entry-points) in the `build` section of the recipe
+- No activation scripts
```
jaimergp: Hm, why? 🤔

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,340 @@
...
+
+All recipes employing `noarch: python` should usually use the `python_min` variable per the following example:
```
jaimergp: Given the language in CFEP 25 ("However, we strongly recommend that maintainers always test their package against the oldest Python version for which it would install in order to ensure compatibility."), I'd say we can voice this more strongly:

```suggestion
All recipes employing `noarch: python` must use the `python_min` variable per the following example:
```

docs/how-to/basics/noarch.mdx
mgorny: As in split the how-to across basic and advanced?

docs/how-to/basics/noarch.mdx
mgorny: Apparently dates back to 9f090507d22fe28c9b2ec24952275e20acdd873f. I can't find an equivalent restriction in the previous text, though. The previous version said:

```
``noarch: python`` can be used to build pure Python packages that:

* Do not perform any Python version specific code translation at install time (i.e. 2to3).

* Have fixed requirements; that is to say no conditional dependencies
  depending on the Python version, or the platform ran. (If you have for example
  ``backports # [py27])`` in the ``run`` section of ``meta.yml``, your package
  can't be noarch, yet.)

* Do not use selectors to ``skip`` building the recipe on a specific platform or
  for a specific version of python (e.g. ``skip: True  # [py2k]``).
```

mgorny: I'll remove it for now. We can revisit if someone figures out why it was there :-).

docs/how-to/basics/noarch.mdx
mgorny: Ok, done.

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,146 @@
...
+If you go that route, you will need to [rerender the feedstock](rerender.md)
+after adding the `conda_build_config.yaml` file.
+
+:::note
+If the minimum Python version supported by the package is older or the same as the minimum version supported by conda-forge (`python_min` in [conda-forge-pinning](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml)), please remove the `python_min` override, so that the global setting (and any subsequent updates to it) take effect.
+:::
+
+### Further notes
+
+1. Adding an explicit `python_min` to your `noarch: python` recipe can be an effective way to ensure the required Python in your package's metadata is enforced at `conda-build` time, as the build will fail if the package's required Python version is newer than `python_min`.
```
jaimergp: I think we have covered this part already.

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,146 @@
...
+2. While `noarch: python` does not work with selectors, it does work with version constraints.
+   `skip: match(python, "<3.12")` can be replaced with `python_min: 3.12` override.
+3. Only `console_scripts` entry points have to be listed in your recipe explicitly, as they require special treatment to create platform-appropriate scripts.
```
jaimergp: Shall we merge this item with the comment in the section above? Not sure why they need to be separate.

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,146 @@
...
+   Other types of entry points do not conflict with `noarch` and therefore do not require extra treatment.
+4. `noarch` is a statement about the package's source code and not its install environment.
+   A package is still considered `noarch` even if one of its dependencies is not available on a given platform.
+   If this is the case, a helpful error message will be printed upon install, describing which dependency couldn't be found.
+   If the dependency is later made available, your package will be installable on that platform without having to make any changes to the feedstock.
+5. If an existing python package qualifies to be converted to a noarch package, you can request the required changes by opening a new issue and including `@conda-forge-admin, please add noarch: python`.
```
jaimergp: Huh, I didn't know about this one. I was going to see whether it's really used these days, but apparently it is. Let's add a link to [this GH Search](https://github.com/search?type=issues&q=org%3Aconda-forge+%22%40conda-forge-admin%2C+please+add+noarch%22) so folks can find examples.

docs/how-to/basics/noarch.mdx
mgorny: I was treating this point as an explanation why we do that (given we don't really explain, just give a requirement/example).

mgorny: Per the comment below, I'll merge this one too.

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,160 @@
+---
+tags: [howto, basic, noarch]
+---
+
+import { RecipeTabs } from "@site/src/components/RecipeTabs";
+
+# Creating noarch packages
+
+`noarch` packages are packages that do not contain architecture specific contents and therefore do not have to be built separately for each architecture.
```
jaimergp:
```suggestion
`noarch` packages are packages that do not ship platform or architecture specific contents and therefore do not have to be built separately for each subdir.
```

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,160 @@
...
+For example, the `tzdata` package provides only data files that are installed into the same layout irrespective of the system used.
+Rather than building a number of roughly the same `linux-64`, `linux-aarch64`, `osx-64`, etc. packages, we can build a single `noarch` package that can be installed on all of these platforms.
+
+Declaring packages as `noarch` reduces the use of shared CI and CDN resources.
+Therefore all packages that qualify to be `noarch` packages should be declared as such.
+
+There are currently two types of `noarch` packages supported:
+- `noarch: python` that is used for Python packages that feature no files specific to Python version
+- `noarch: generic` that is used for non-Python `noarch` packages
```
jaimergp:
```suggestion
- `noarch: generic` that is used for all other cases (e.g. R, Perl, data-only packages)
```

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,160 @@
...
+- The package contents must be invariant across Python versions and platforms.
+  - This means that the package must not compile any extensions.
+    For packages with `abi3` extensions, see [ABI3 packages](/docs/maintainer/knowledge_base/#packages-with-abi3-extensions).
+  - Typically, the package must not rely on platform-specific build scripts, unless they produce the same results.
+- No post-link or pre-link or pre-unlink scripts may be used.
+- The recipe must not use selectors.
+  <!-- TODO: discuss repodata v3 conditional dependencies when ready -->
+  - Requirements that are version-restricted upstream can be listed unconditionally in the recipe instead.
+  - If the specific requirement is incompatible with a newer Python version, the package will built successfully.
+    Conda will display an appropriate error if one attempts to install this package along with the newer Python version.
```
jaimergp:
```suggestion
    The package manager will display an appropriate error if one attempts to install this package along with the newer Python version.
```

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,160 @@
...
+    However, when the dependency is rebuilt for the new Python version, your package will become installable immediately, with no need to build a new version.
+  - Instead of `skip: match(python, ...)`, `python_min` needs to be overridden, see [overriding the minimum Python version](#overriding-the-minimum-python-version).
+  - The advanced how-to explains how to express [OS-specific dependencies](../advanced/noarch.mdx#noarch-packages-with-os-specific-dependencies).
+- `scripts` argument in `setup.py` must not be used.
+  - It is recommended to replace explicit scripts with entry points upstream.
+- If `console_scripts` entry points are used, they must be [listed](https://conda.io/projects/conda-build/en/stable/resources/define-metadata.html#python-entry-points) in the `build` section of the recipe.
```
jaimergp: Can we mention the field name `entry_points`? Is it the same for v1?

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,160 @@
...
+  This is necessary for conda to generate the correct platform-appropriate scripts.
```
jaimergp:
```suggestion
  This is necessary for the package manager to generate the correct platform-appropriate scripts.
```

docs/how-to/advanced/noarch.mdx
```diff
@@ -0,0 +1,202 @@
...
+relies on three concepts:
+
+1. [Virtual packages](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html).
+   Prefixed with a double underscore, they are used by conda to represent system properties as
+   constraints for the solver at install-time. We will use `__linux`, `__win` or `__osx`,
+   which are only present when the running platform is Linux, Windows, or MacOS, respectively.
+   `__unix` is present in both Linux and MacOS. Note that this feature is **only fully available
+   on conda 4.10 or above**.
+2. `conda-forge.yml`'s [noarch_platforms](/docs/maintainer/conda_forge_yml.mdx#noarch-platforms) option.
+3. **conda-build 3.25.0 or above** changing the build hash depending on virtual packages used.
```
jaimergp: I think this one is no longer necessary to indicate, it's been a while we are on more recent versions.

docs/how-to/advanced/noarch.mdx
```diff
@@ -0,0 +1,202 @@
...
+    - if: win
+      then: windows-only-dependency
+```
+
+</RecipeTabs>
+
+With this recipe, the build matrix will include separate builds for every supported platform, and for every supported Python version on every platform.
+All the built package will have the same contents, and only `win-64` packages will have different dependencies (though they still will be duplicated across all Python versions).
+
+We can get it down to two packages if replace it with this other approach!
```
jaimergp:
```suggestion
We can get it down to two packages if replace it with this other approach:
```

I was too happy when I wrote the original text, apparently 😂

docs/how-to/advanced/noarch.mdx
```diff
@@ -0,0 +1,202 @@
...
+      then:
+        - __unix
+      else:
+        - __win
+        - windows-only-dependency
+```
+
+</RecipeTabs>
+
+Do not forget to specify the platform virtual packages with their selectors!
```
jaimergp:
```suggestion
Do not forget to specify the platform virtual packages with their selectors.
```

docs/how-to/advanced/noarch.mdx
```diff
@@ -0,0 +1,202 @@
...
+the `noarch_platforms` option in `conda-forge.yml`:
+
+```yaml title="conda-forge.yml"
+noarch_platforms:
+  - linux_64
+  - win_64
+```
+
+This will provide two runners per package! Perfect! All these changes require a
+feedstock rerender to be applied. See [Rerendering feedstocks](../basics/rerender.md).
```
jaimergp:
```suggestion
Once [rerendered](../basics/rerender.md), this will result in two CI jobs.
```

jaimergp (commented):
This rewrite is very refreshing, thanks. Added some comment to finish some details.

docs/how-to/basics/noarch.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+### Converting existing packages to noarch
+
+If an existing python package qualifies to be converted to a noarch package, you can request the required changes by opening a new issue and including `@conda-forge-admin, please add noarch: python` (see [examples](https://github.com/search?type=issues&q=org%3Aconda-forge+%22%40conda-forge-admin%2C+please+add+noarch%22)).
+
+## noarch: generic
+
+:::info[Todo]
+
+add some information on r packages which make heavy use of `noarch: generic`
```
jaimergp, line 173: Can we add a link to a GH search featuring generic packages as of today? Just as a minimal example. e.g. https://github.com/search?type=code&q=org%3Aconda-forge+path%3Arecipe+%22noarch%3A+generic%22+language%3AYAML&l=YAML

mgorny: To be honest, I originally planned to write this part too, but I think the PR grew too big, so I'll defer it to another.

jaimergp (approved):
LGTM, thanks for all the iterations! A note for the future: some how-to's are becoming longer and venturing into "explanation"-like content, so at some point we may need to rethink that. e.g. what a noarch package is can be an explanation that is separate from the actual set of instruction to convert a recipe into `noarch: python`. We can defer that once we have a bigger corpus of precedents, though.

PR #2934, opened 2026-09-10
Start expanding `noarch: generic`: requirements and R

mgorny:
PR Checklist:

I'm trying to figure out a "best reference" here, as a combination of what I've found in real recipes, `grayskull` and `rattler-build generate` output.

For R recipes, apparently v0 format also supports using `imports:` but according to my testing, it works only on Linux and fails weirdly on Windows. Either way, it looks like a bad solution to recommend.


mgorny:
Let's review it as-is. We can add more types of `noarch: generic` packages later.

